PR #56, opened 2020-09-23
fix: Adding read the docs specifics

abhilash-kumar-nair:
Read the docs configs
@efredriksson-modelon  - Poetry is not supported by sphinx hence the need for requirements.txt.


efredriksson-modelon (approved):
👍

PR #61, opened 2020-09-25
fix: delete invalid api keys to let user reenter new ones

jacobtaxen:
@abhilash-kumar-nair

### DIFF
--- a/modelon/impact/client/client.py
+++ b/modelon/impact/client/client.py
@@ -63,6 +63,7 @@ class Client:
         self.uri = modelon.impact.client.sal.service.URI(url)
         self._sal = modelon.impact.client.sal.service.Service(self.uri, context)
         self._credentials = credential_manager
+        self._api_key = None
 
         self._validate_compatible_api_version()
 
@@ -73,7 +74,7 @@ class Client:
                 logger.warning(
                     "The provided API key is not valid, please enter a new key"
                 )
-                self.api_key = self._credentials.get_key_from_prompt()
+                self._api_key = self._credentials.get_key_from_prompt()
                 self._authenticate_against_api(interactive)
             else:
                 raise
@@ -97,22 +98,23 @@ class Client:
 
     def _authenticate_against_api(self, interactive):
 
-        api_key = self._credentials.get_key(interactive=interactive)
+        if self._api_key is None:
+            self._api_key = self._credentials.get_key(interactive=interactive)
 
-        if api_key is None:
+        if self._api_key is None:
             logger.warning(
                 "No API key could be found, will log in as anonymous user. "
                 "Permissions may be limited"
             )
             login_data = {}
         else:
-            login_data = {"secretKey": api_key}
+            login_data = {"secretKey": self._api_key}
 
         self._sal.api_login(login_data)
         if interactive:
             # Save the api_key for next time if
             # running interactively and login was successfuly
-            self._credentials.write_key_to_file(api_key)
+            self._credentials.write_key_to_file(self._api_key)
 
     def get_workspace(self, workspace_id):
         """
--- a/modelon/impact/client/client.py
+++ b/modelon/impact/client/client.py
@@ -2,6 +2,7 @@
 import logging
 import modelon.impact.client.configuration
 import modelon.impact.client.entities
+import modelon.impact.client.exceptions
 import modelon.impact.client.sal.service
 import modelon.impact.client.sal.exceptions
 import modelon.impact.client.credential_manager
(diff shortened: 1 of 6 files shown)


modelon/impact/client/client.py
```diff
@@ -71,7 +72,16 @@ def _authenticate_against_api(self, interactive):
         else:
             login_data = {"secretKey": self.api_key}
 
-        self._sal.api_login(login_data)
+        try:
+            self._sal.api_login(login_data)
+        except modelon.impact.client.sal.exceptions.HTTPError as error:
+            if interactive:
+                self._credentials.delete_key()
```
efredriksson-modelon, line 79: It does not seem ok to me deleting the users key if something goes wrong. Could be very destructive if this is not what the user wants (especially if the only place they keep the key is in the file).

It might be a good idea to move this one level up:
```
try:
  self._authenticate_against_api(interactive)
except modelon.impact.client.sal.exceptions.HTTPError as error:
  if interactive:
    logger.warning('Could not login with configured API key, re-trying with input from prompt')
    self.api_key = self._credentials.get_key_from_prompt()
    self._authenticate_against_api(interactive)
  else:
    raise
```

modelon/impact/client/credential_manager.py
```diff
@@ -34,3 +34,6 @@ def get_key(self, interactive=False):
             return api_key
 
         return self.get_key_from_prompt()
+
+    def delete_key(self):
```
efredriksson-modelon, line 38: delete_key_from_file

efredriksson-modelon (changes requested):
Nice to see improvements to the api login 👍 . I am not entierly sure on the problem this solves. This is just for the case when the API key as no longer valid and the user is running in interactive mode? I gave a suggestion for an alternative solution that should be a bit less risky.

modelon/impact/client/client.py
```diff
@@ -71,7 +72,16 @@ def _authenticate_against_api(self, interactive):
...
+            raise modelon.impact.client.exceptions.InvalidAPIKeyError(
```
efredriksson-modelon, line 80: We should probably skip raising this specific error from the generic http error. We should do something with error codes for more precise errors in the future, but for now we should not try to make to many assumptions that risk introducing bugs/confusion.

efredriksson-modelon (changes requested):

tests/impact/client/test_client.py
```diff
@@ -148,16 +150,18 @@ def test_client_login_fail_interactive_dont_save_key(_, login_fails):
...
     cred_manager = unittest.mock.MagicMock()
+    cred_manager.get_key_from_prompt.return_value = 'test_client_login_still_fails'
+
     pytest.raises(
-        modelon.impact.client.exceptions.InvalidAPIKeyError,
+        modelon.impact.client.sal.exceptions.HTTPError,
         modelon.impact.client.Client,
         url=login_fails.url,
         context=login_fails.context,
         credentail_manager=cred_manager,
```
abhilash-kumar-nair, line 162: Could you correct the spelling? 😥 
credentail_manager --> credential_manager

abhilash-kumar-nair (changes requested):
Please re-base! :)

efredriksson-modelon (approved):
🥇 (the test just needs to be rewritten after rebase)

abhilash-kumar-nair (approved):
Neat! 🥇 💯

modelon/impact/client/client.py
```diff
@@ -46,29 +47,36 @@ class Client:
...
-        self._authenticate_against_api(interactive)
+
+        try:
+            self._authenticate_against_api(interactive)
+        except modelon.impact.client.sal.exceptions.HTTPError:
+            if interactive:
+                logger.warning(
+                    "The provided API key is not valid, please enter a new key"
+                )
+                self.api_key = self._credentials.get_key_from_prompt()
```
efredriksson-modelon, line 76: I think this don't work anymore because, `_authenticate_against_api` does not read from self.api_key anymore. Could probably add back so it reads from self.api_key and set self.api_key to None at the start of the constructor.

efredriksson-modelon (changes requested):

modelon/impact/client/client.py
efredriksson-modelon: The test goes through because it does not check that api/login is called twich with different API keys, only that get_key and get_key_from_prompt is called. Probably better if the test check what api/login is called with as that is what is important.

efredriksson-modelon (approved):
🍉

PR #64, opened 2020-09-28
fix: Avoid users to hard-code api keys

abhilash-kumar-nair:
@jacobtaxen


efredriksson-modelon (approved):
Thanks for fixing this 👍

PR #66, opened 2020-10-14
feat: Using the case trajectory api to fetch case trajectories

abhilash-kumar-nair:
@jacobtaxen


efredriksson-modelon (approved):
Nice to have this fixed! 🍉

PR #67, opened 2020-10-15
feat: Added checks to use cached FMU if available

abhilash-kumar-nair:
**How to test-**
**Basic Tests**
1. Execute the below code once to create an FMU. 
2. Execute the code again. This time model should not be recompiled(see logs)
3. Change the option generate_html_diagnostics to true. Execute the code again, the Model should recompile now.
 
**Coupling with UI test.**
1.  Compile a model in UI.
2. Change any param in modeling mode
3. Simulate the model using client. Check if parameter modifiers have been correctly applied

```

workspace = client.create_workspace("Hello6")

# Choose analysis type

# Compile model
model = workspace.get_model("Workspace.PID_Controller")
fmu = model.compile(
    compiler_options=dynamic.get_compiler_options(), compiler_log_level="w",
).wait()
print(fmu.modifiers)

# # Execute experiment
experiment_definition = fmu.new_experiment_definition(dynamic).with_modifiers(
    {'PI.k': 120}
)
case = exp.get_case('case_1')
res = case.get_trajectories()
k = res['PI.k']
```

@jacobtaxen

### DIFF
--- a/modelon/impact/client/entities.py
+++ b/modelon/impact/client/entities.py
@@ -1,5 +1,6 @@
 import os
 import tempfile
+import logging
 from modelon.impact.client import operations
 
 from modelon.impact.client.experiment_definition import SimpleExperimentDefinition
@@ -7,6 +8,8 @@ from collections.abc import Mapping
 from modelon.impact.client.options import ExecutionOptions
 from modelon.impact.client import exceptions
 from enum import Enum
+
+logger = logging.getLogger(__name__)
 
 
 def _assert_successful_operation(is_successful, operation_name="Operation"):
@@ -677,6 +680,7 @@ class Model:
         fmi_target="me",
         fmi_version="2.0",
         platform="auto",
+        force_compilation=False,
     ):
         """Compiles the model to an FMU.
         Returns an modelon.impact.client.operations.ModelExecutableOperation class
@@ -707,6 +711,9 @@ class Model:
             platform --
                 Platform for FMU binary. Supported values are "auto", "win64", "win32"
                 or "linux64". Default: 'auto'.
+
+            force_compilation --
+                Force a model compilation.
 
         Returns:
 
@@ -751,9 +758,26 @@ class Model:
                 "platform": platform,
             }
         }
+        if not force_compilation:
+            fmu_id, modifiers = self._model_exe_sal.fmu_setup(
+                self._workspace_id, body, True
+            )
+            if fmu_id:
+                return operations.CachedModelExecutableOperation(
+                    self._workspace_id,
+                    fmu_id,
+                    self._workspace_sal,
+                    self._model_exe_sal,
+                    None,
+                    modifiers,
+                )
+
+        # No cached FMU, setup up a new one
+        fmu_id, _ = self._model_exe_sal.fmu_setup(self._workspace_id, body, False)
+
         return operations.ModelExecutableOperation(
             self._workspace_id,
-            self._model_exe_sal.compile_model(self._workspace_id, body),
+            self._model_exe_sal.compile_model(self._workspace_id, fmu_id),
             self._workspace_sal,
             self._model_exe_sal,
         )
@@ -771,12 +795,14 @@ class ModelExecutable:
         workspace_service=None,
         model_exe_service=None,
         info=None,
+        modifiers=None,
     ):
         self._workspace_id = workspace_id
         self._fmu_id = fmu_id
         self._workspace_sal = workspace_service
         self._model_exe_sal = model_exe_service
         self._info = info
+        self._modifiers = modifiers
 
     def __repr__(self):
         return f"FMU with id '{self._fmu_id}'"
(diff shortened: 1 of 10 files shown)


efredriksson-modelon (changes requested):
This one needs a bit more work. It does not support handling the case where there is an FMU cached but there are some additional parameters that needs to be applied to the FMU. It is also not possible force a recompilation, like `model.compile(..., force_compilaton=True)`.

We should also not start a compilation if we find a cache, which is inefficient. Given that we make sure service has no logic in it I think the entity compile method should look something like:

```
if not force_compilation:
  fmu_id, parameters = self.fmu_service.setup_fmu(...., True)
  if fmu_id is not None:
     return CachedModelExecutableOperator(fmu_id, parameters)

# No cached FMU, setup up a new one
fmu_id, _ = self.fmu_service.setup_fmu(...., False)
self.fmu_service.compile_fmu(..., fmu_id)
return ModelExecuatableOperator(fmu_id)
```

abhilash-kumar-nair:
> This one needs a bit more work. It does not support handling the case where there is an FMU cached but there are some additional parameters that needs to be applied to the FMU. It is also not possible force a recompilation, like `model.compile(..., force_compilaton=True)`.
> 
> We should also not start a compilation if we find a cache, which is inefficient. Given that we make sure service has no logic in it I think the entity compile method should look something like:
> 
> ```
> if not force_compilation:
>   fmu_id, parameters = self.fmu_service.setup_fmu(...., True)
>   if fmu_id is not None:
>      return CachedModelExecutableOperator(fmu_id, parameters)
> 
> # No cached FMU, setup up a new one
> fmu_id, _ = self.fmu_service.setup_fmu(...., False)
> self.fmu_service.compile_fmu(..., fmu_id)
> return ModelExecuatableOperator(fmu_id)
> ```

@efredriksson-modelon , Thanks for the review! I have made the suggested changes. Could please test it out? I have added an example script with the description. Also, do you think we require any specific documentation updates for the changes in this PR?

tests/impact/client/fixtures.py
```diff
@@ -325,6 +325,21 @@ def experiment_create(sem_ver_check, mock_server_base):
...
+def no_cached_fmu_id(mock_server_base):
+    json = {
+        "id": None,
+        "parameters": {},
+    }
+
+    return with_json_route(
+        mock_server_base,
+        'POST',
+        'api/workspaces/WS/model-executables' '?getCached=true',
```
efredriksson-modelon, line 338: minor formatting: Seems to be 2 strings with a space between them.

efredriksson-modelon (changes requested):
Much better 👍 Just a few more things to iron out.

modelon/impact/client/entities.py
```diff
@@ -789,6 +815,28 @@ def id(self):
...
+                    "compiler_log_level": input_data['compiler_log_level'],
+                    "fmi_target": input_data['fmi_target'],
+                    "fmi_version": input_data['fmi_version'],
+                    "platform": input_data['platform'],
+                }
+            }
+            _, self._modifiers = self._model_exe_sal.fmu_setup(
+                self._workspace_id, body, True
+            )
+        return self._modifiers
```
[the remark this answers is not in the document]
abhilash-kumar-nair: Ah you are right! 💡

modelon/impact/client/operations.py
```diff
@@ -201,17 +194,100 @@ def cancel(self):
...
+
+        Example::
+
+            model.compile(options).status()
+        """
+        return Status(
+            self._model_exe_sal.compile_status(self._workspace_id, self._fmu_id)[
+                "status"
+            ]
+        )
```
efredriksson-modelon, line 276: `return Status.DONE`

modelon/impact/client/operations.py
```diff
@@ -201,17 +194,100 @@ def cancel(self):
         self._model_exe_sal.compile_cancel(self._workspace_id, self._fmu_id)
 
 
-class ExperimentOperation(Operation):
+class CachedModelExecutableOperation(Operation):
```
efredriksson-modelon, line 197: I think we also should overwrite the `wait` method to make things go fast, I am thinking something like:
            
```
if self.status() != status:
     raise exceptions.OperationTimeOutError(f"The operation '{self.name}' have the status '{self.status().name}', it will never get the status '{status.name}'")

logger.info("Found cached FMU which will be used")
return self.data()
```

efredriksson-modelon (changes requested):

modelon/impact/client/operations.py
abhilash-kumar-nair: Good Idea! But I was thinking of just adding a special case to the base class **wait()** function instead of overwriting it. If we overwrite it in the child class, we would have to keep a dummy 'timeout' parameter,which seemed a bit ugly to me?
I was thinking of something like this :-
```
        if self.name == "Looking for cached FMU":
            if self.status() != status:
                raise exceptions.OperationTimeOutError(
                    f"The operation '{self.name}' has the status '{self.status().name}', "
                    f"it will never get the status '{status.name}'!"
                )

            logger.info("Cached FMU found! Using the cached FMU!")
            return self.data()
```

efredriksson-modelon: A base class should never need to know about implementation details in classes that inherents from it. That the timeout parameter would have no effect for the overriden method seems ok to me.

efredriksson-modelon (approved):
👍

PR #72, opened 2020-12-23
fix: Corrected the upload_workspace api

abhilash-kumar-nair:
This PR fixes the upload workspace API.

### DIFF
--- a/modelon/impact/client/client.py
+++ b/modelon/impact/client/client.py
@@ -47,11 +47,7 @@ class Client:
     _SUPPORTED_VERSION_RANGE = ">=1.8.0,<2.0.0"
 
     def __init__(
-        self,
-        url=None,
-        interactive=None,
-        credential_manager=None,
-        context=None,
+        self, url=None, interactive=None, credential_manager=None, context=None,
     ):
         if url is None:
             url = modelon.impact.client.configuration.get_client_url()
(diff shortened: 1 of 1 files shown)


jacobtaxen (approved):
Looks good and works fine

modelon/impact/client/sal/service.py
```diff
@@ -558,6 +567,26 @@ def data(self):
         return self._resp_obj.content
 
 
+class FormDataResponse(Response):
```
efredriksson-modelon, line 570: This class is not used anywhere?

modelon/impact/client/sal/service.py
```diff
@@ -341,6 +341,10 @@ def post_json(self, url, body=None, files=None):
         request = RequestJSON(self._context, "POST", url, body, files)
         return request.execute().data
 
+    def post_form_data(self, url):
```
efredriksson-modelon, line 344: This method is not used anywhere?

efredriksson-modelon (changes requested):
I am a little confused as the 'Form' classes/methods are not used.

abhilash-kumar-nair:
> I am a little confused as the 'Form' classes/methods are not used.

@efredriksson-modelon , You are right! The response body would be still be a JSON, so the only issue was  the additional workspace Id argument . I have made the correction now. Thanks!

efredriksson-modelon (approved):
👍

PR #73, opened 2021-01-04
fix: Removed kwargs argument as we only need to support 'qualified' v…

abhilash-kumar-nair:
Code snippet to try. @efredriksson-modelon  -->

```

workspace = client.create_workspace("Hello1")

# Choose analysis type

# Compile model
fmu = model.compile(compiler_options=dynamic.get_compiler_options()).wait()
# print(fmu.id)

# # Execute experiment
experiment_definition = fmu.new_experiment_definition(dynamic).with_modifiers(
    driveAngle=40
)

```

### DIFF
--- a/modelon/impact/client/experiment_definition.py
+++ b/modelon/impact/client/experiment_definition.py
@@ -1,7 +1,11 @@
+import logging
+
 from modelon.impact.client import entities
 from abc import ABC, abstractmethod
 from modelon.impact.client.options import ExecutionOptions
 from modelon.impact.client import exceptions
+
+logger = logging.getLogger(__name__)
 
 
 def _get_options(default_options, options):
@@ -262,13 +266,7 @@ class SimpleFMUExperimentDefinition(BaseExperimentDefinition):
         Parameters:
 
             modifiers --
-                A dictionary of variable modifiers. Could be used if
-                modifiers keys conflict with python identifiers or keywords.
-                Default: None.
-
-            modifiers_kwargs --
-                A keyworded, variable-length argument list of variable
-                modifiers.
+                A dictionary of variable modifiers.
 
         Example::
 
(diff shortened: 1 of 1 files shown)


efredriksson-modelon (changes requested):
This will be a breaking change so unless there is a mayor gain for this change, I don't think we should do this.

If there is a usability problem, we should update the documentation. If we really don't like this form of giving modifiers, we could remove all traces of it from the documentation but keep the implementation. A middle way could be to ensure the examples all use the preferred way, so the reader only need to take in one way of specifying modifiers.

abhilash-kumar-nair:
> This will be a breaking change so unless there is a mayor gain for this change, I don't think we should do this.
> 
> If there is a usability problem, we should update the documentation. If we really don't like this form of giving modifiers, we could remove all traces of it from the documentation but keep the implementation. A middle way could be to ensure the examples all use the preferred way, so the reader only need to take in one way of specifying modifiers.

@efredriksson-modelon, I have removed **kwargs** from the docs and also added a deprecation warning. However I have still kept the usage in the test,  we have for the 'with_modifier' function. I feel we should have them tested until we finally deprecate them in a future release. What do you think?

modelon/impact/client/experiment_definition.py
```diff
@@ -1,8 +1,13 @@
+import logging
+
 from modelon.impact.client import entities
 from abc import ABC, abstractmethod
 from modelon.impact.client.options import ExecutionOptions
 from modelon.impact.client import exceptions
 
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
```
efredriksson-modelon, line 9: remove

modelon/impact/client/experiment_definition.py
```diff
@@ -240,6 +239,12 @@ def with_modifiers(self, modifiers=None, **modifiers_kwargs):
                 custom_function).with_modifiers({'inertia1.J': Choices(0.1, 0.9),
                 'inertia2.J': Range(0.1, 0.5, 3)})
         """
+        if modifiers_kwargs:
+            logger.info(
```
efredriksson-modelon, line 243: info -> warning

modelon/impact/client/experiment_definition.py
```diff
@@ -468,6 +467,12 @@ def with_modifiers(self, modifiers=None, **modifiers_kwargs):
             ).with_modifiers({'PI.k': 40})
 
         """
+        if modifiers_kwargs:
+            logger.info(
```
efredriksson-modelon, line 471: info -> warning

efredriksson-modelon (changes requested):
We also need to update the sphinx documentation.

efredriksson-modelon (approved):
👍

PR #76, opened 2021-01-07
feat: Added support for download result artifact

abhilash-kumar-nair:


modelon/impact/client/sal/service.py
```diff
@@ -264,6 +264,17 @@ def case_trajectories_get(self, workspace_id, experiment_id, case_id, variables)
...
 
+    def case_artifact_get(self, workspace_id, experiment_id, case_id, artifact_id):
+        url = (
+            self._base_uri
+            / f"api/workspaces/{workspace_id}/experiments/{experiment_id}/cases/"
+            f"{case_id}/custom-artifacts/{artifact_id}"
+        ).resolve()
+        result, headers = self._http_client.get_octet_stream(url)
+        d = headers["content-disposition"]
+        file_name = re.findall("filename=(.+)", d)[0].strip('"')
```
efredriksson-modelon, line 275: These lines are now repeated twice. I think we should factor it out to a function like _get_filename_from_headers. 

Another approach that could be nice is to change `get_octet_stream` to `get_octet_response` which returns `OctetStreamResponse` back instead. Then you could change `data` to `stream` and also add `filename`. So this code would read something like:

```
resp = self._http_client.get_octet_response(url)
return resp.stream, resp.filename
```

efredriksson-modelon (changes requested):
Seems reasonable 👍

I think some refactoring is needed to keep it manageble.

modelon/impact/client/sal/service.py
abhilash-kumar-nair: Good idea! I have made the changes! Could you please take another look? Thanks!

efredriksson-modelon (approved):
🍉

PR #77, opened 2021-01-12
feat: Added support for modelica_class name based workflows

abhilash-kumar-nair:
You should be able to do :-

```
from modelon.impact.client import Client, Range, SimpleExperimentExtension

workspace = client.create_workspace("Test")

# Choose analysis type

# Compile model
experiment_definition = model.new_experiment_definition(dynamic).with_cases(
    [{'PI.k': 20}, {'PI.k': 30}]
)
experiment_definition = model.new_experiment_definition(dynamic).with_extensions(
    [
        SimpleExperimentExtension(
            parameter_modifiers={'start_time': 0.0, 'final_time': 2.0}
        ).with_modifiers({'PI.k': 20})
    ]
)
experiment_definition = model.new_experiment_definition(dynamic).with_modifiers(
    {'PI.yMax': Range(12, 15, 4)}
)
# Execute experiment
print(experiment_definition.to_dict())

```

**Stuff to do**
1. Add  RTD documentation
2.  Add tests.

### DIFF
--- a/modelon/impact/client/entities.py
+++ b/modelon/impact/client/entities.py
@@ -3,7 +3,11 @@ import tempfile
 import logging
 from modelon.impact.client import operations
 
-from modelon.impact.client.experiment_definition import SimpleExperimentDefinition
+from modelon.impact.client.experiment_definition import (
+    SimpleModelicaExperimentDefinition,
+    SimpleFMUExperimentDefinition,
+    assert_valid_args,
+)
 from collections.abc import Mapping
 from modelon.impact.client.options import ExecutionOptions
 from modelon.impact.client import exceptions
@@ -58,21 +62,6 @@ def _create_result_dict(variables, workspace_id, exp_id, case_id, exp_sal):
         variable: response[i][case_index - 1] for i, variable in enumerate(variables)
     }
     return data
-
-
-def _assert_valid_args(compiler_options, runtime_options):
-    if not isinstance(compiler_options, (ExecutionOptions, dict)):
-        raise TypeError(
-            "Compiler options object must either be a dictionary or an "
-            "instance of modelon.impact.client.options.ExecutionOptions class!"
-        )
-    if runtime_options is not None and not isinstance(
-        runtime_options, (ExecutionOptions, dict)
-    ):
-        raise TypeError(
-            "Runtime options object must either be a dictionary or an "
-            "instance of modelon.impact.client.options.ExecutionOptions class!"
-        )
 
 
 class ModelExecutableStatus(Enum):
@@ -426,7 +415,9 @@ class Workspace:
 
             definition --
                 An parametrized experiment definition class of type
-                modelon.impact.client.experiment_definition.SimpleExperimentDefinition.
+                modelon.impact.client.experiment_definition.SimpleModelicaExperimentDefinition
+                or
+                modelon.impact.client.experiment_definition.SimpleFMUExperimentDefinition.
 
         Returns:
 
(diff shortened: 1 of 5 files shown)


modelon/impact/client/entities.py
```diff
@@ -434,13 +425,16 @@ def create_experiment(self, definition):
 
             workspace.create_experiment(definition)
         """
-        if isinstance(definition, SimpleExperimentDefinition):
+        if isinstance(definition, SimpleFMUExperimentDefinition):
+            definition = definition.to_dict()
+        elif isinstance(definition, SimpleModelicaExperimentDefinition):
```
efredriksson-modelon, line 430: Use `BaseExperimentDefinition` and add 'to_dict' as an abstract method on it.

modelon/impact/client/experiment_definition.py
```diff
@@ -366,6 +395,303 @@ def to_dict(self):
...
+            if isinstance(compiler_options, dict)
+            else {}
+        )
+        self._runtime_options = (
+            dict(runtime_options)
+            if isinstance(runtime_options, ExecutionOptions)
+            else runtime_options
+            if isinstance(runtime_options, dict)
+            else {}
+        )
```
efredriksson-modelon, line 495: These options should also get default options from the custom function if they are None. After that this construct will be repeated 6 times in this file and some refactoring should be done. Also make sure the documentation is updated to mention what None means.

modelon/impact/client/experiment_definition.py
```diff
@@ -366,6 +395,303 @@ def to_dict(self):
...
+        """Sets the modifiers parameters for an experiment.
+
+        Parameters:
+
+            modifiers --
+                A dictionary of variable modifiers. Could be used if
+                modifiers keys conflict with python identifiers or keywords.
+                Default: None.
+
+            modifiers_kwargs --
```
efredriksson-modelon, line 533: remove

efredriksson-modelon (changes requested):
Over-all good 👍

modelon/impact/client/experiment_definition.py
```diff
@@ -217,20 +234,13 @@ def __init__(
...
         self.custom_function = custom_function
-        self._solver_options = (
-            dict(custom_function.get_solver_options())
-            if solver_options is None
-            else dict(solver_options)
-            if isinstance(solver_options, ExecutionOptions)
-            else solver_options
+        self._solver_options = _parse_options(
+            dict(custom_function.get_solver_options()), solver_options
         )
```
efredriksson-modelon, line 239: I think the first argument should just be the funtion which can be used for getting the default `custom_function.get_solver_options`. The rest can be done in the function. The main reason for this is so to only call the server asking for options if we are going to use them. Now we will do the call regardless.

For when handling the case where you want an empty by default you can send in `dict`.

modelon/impact/client/experiment_definition.py
```diff
@@ -4,10 +4,30 @@
 from modelon.impact.client import exceptions
 
 
-def _assert_valid_args(fmu, custom_function, solver_options, simulation_options):
-    if not isinstance(fmu, entities.ModelExecutable):
-        raise TypeError("Fmu must be an instance of ModelExecutable class!")
-    if not isinstance(custom_function, entities.CustomFunction):
+def _parse_options(default_options, options):
```
efredriksson-modelon, line 7: The name of this function is a bit missleading as it does not parse anything. I think just `_get_options` is fine.

efredriksson-modelon (changes requested):
Just one more thing.

efredriksson-modelon (approved):
👍

PR #79, opened 2021-01-28
fix: Remove trailing commas when single values are given in Choices o…

abhilash-kumar-nair:


modelon/impact/client/experiment_definition.py
```diff
@@ -159,7 +159,7 @@ class Choices(Operator):
     """
 
     def __init__(self, *values):
-        self.values = values
+        self.values = str(values).replace(',)', ')')
```
efredriksson-modelon, line 162: It would be cleaner to have the conversion to string in the __str__ method as that is what it does. Also, for this case where the `tuple` `__str__` method does not work for us (due to ',' if single element) it is better to implement it ourself than trying to patch it up. The motivation is to get explicit code that does what you expect.

You could do it like this:
f"choices({', '.join(map(str, self.values))})"

efredriksson-modelon (changes requested):
Good find 👑 

Can you also add a test case for what was not working previously?

modelon/impact/client/experiment_definition.py
abhilash-kumar-nair: Thanks for reviewing! I agree this makes more sense!

efredriksson-modelon (approved):
👍

PR #83, opened 2021-02-08
fix: correct docs on model.new_experiment_definition

jacobtaxen:
Following the previous documentation would give you a very hard to decipher error message. 

I have also added a ,*, in the model.new_experiment_definition signature to enforce the keywords, so that you have to set the arguments as keyword args. 

(Also a lot of trailing whitespaces have been removed)


efredriksson-modelon (changes requested):
This seems like a reasonable way to reduce this kind of problems 👍 . For symmetry we should make the samme change to the SimpleModelicaExperimentDefinition constructor.

PR #84, opened 2021-02-08
fix: change docs to match current signature of client.upload_workspace

jacobtaxen:
The docs claimed that you can specify a name when uploading a workspace, which you can't anymore. The current upload_workspace takes a path to a zip-file as its only argument


efredriksson-modelon (approved):
🍉

PR #85, opened 2021-02-09
fix: Updated the metadata property to use the new POST /steady-state-metadata api

abhilash-kumar-nair:


efredriksson-modelon (approved):
🍉

PR #86, opened 2021-02-09
fix: Made fmu.modifier method private

abhilash-kumar-nair:
Made the modifiers private as this maybe a bit misleading for the user to understand what it means. Also since it can only contain a value if the change is made in the client, it makes it of little use for it to be public


efredriksson-modelon (approved):
Good initiative to hide it, we can expose it in the future 👍

PR #90, opened 2021-03-25
fix: clarification on 'interactive' argument in client docstring

efredriksson-modelon:


modelon/impact/client/client.py
```diff
@@ -25,9 +25,19 @@ class Client:
...
 
         interactive --
-            True, if running an interactive python shell. Defaults to the value
-            specified by env variable 'MODELON_IMPACT_CLIENT_INTERACTIVE' if set else
-            False.
+            If True the client will prompt for an API key if no other login information
+            can be found. An API key entered for this prompt will be saved and re-used
+            next time the Client is instantiated. If False no prompt will be given if
+            no other login information can be found and login will be done as if no
+            API key was given (anonymous login).
```
maraberg, line 32: Very good! 

One minor thing could be added: Where is key saved?

maraberg (commented):
I think it is ok, 

I think that mentioning where key is stored could be valuable. There might be reasons for not doing this I am not aware of though.

maraberg (approved):
Looks good!

modelon/impact/client/client.py
```diff
@@ -25,9 +25,19 @@ class Client:
             'http://localhost:8080/'.
 
         interactive --
-            True, if running an interactive python shell. Defaults to the value
-            specified by env variable 'MODELON_IMPACT_CLIENT_INTERACTIVE' if set else
-            False.
+            If True the client will prompt for an API key if no other login information
+            can be found. An API key entered for this prompt will be saved to disc
```
jacobtaxen, line 29: disc -> disk

jacobtaxen (approved):
Looks good to me

PR #92, opened 2021-07-02
fix: exmple creating set of all FMUs does now works

efredriksson-modelon:
The code example in the documentation:
```
# Getting a set of simulated FMU objects from the cases
fmus = set(case.get_fmu() for case in exp.get_cases())
```
did not work because the fmu entity was not hashable.


abhilash-kumar-nair (approved):
Good catch! 👍 Could you please rebase to beta? Thanks!

PR #105, opened 2021-08-26
More review fixes

efredriksson-modelon:
This changes so case data is not fetched excessively. It also makes things more deterministic by only fetching case data when a case object is created.

### DIFF
--- a/modelon/impact/client/entities.py
+++ b/modelon/impact/client/entities.py
@@ -1692,9 +1692,6 @@ class Case:
         """Experiment id"""
         return self._exp_id
 
-    def _get_info(self):
-        return self._info
-
     @property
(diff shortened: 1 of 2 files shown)


modelon/impact/client/entities.py
```diff
@@ -1694,22 +1693,18 @@ def experiment_id(self):
         return self._exp_id
 
     def _get_info(self):
```
abhilash-kumar-nair, line 1695: Can we remove this method? This was never intended to be used by the user right?

efredriksson-modelon: Fixed!

abhilash-kumar-nair (approved):
💯

PR #108, opened 2021-09-01
Import external result

abhilash-kumar-nair:
Code to test - 
```

client = Client(url="http://127.0.0.1:8080/")
workspace = client.create_workspace('Test')
res_path = r"C:\Users\AbhilashKumar\Downloads\Result 1.mat"
a = workspace.upload_result(res_path ).wait()
a.delete()
```
@andrewwren-modelon I have a few questions on the API end - 
1. Is it possible to upload any file through this API, not just mat? What I observed that any file below 100 MB can be uploaded and the files are renamed to result.mat. 
2. Does the API POST/uploads/results hold the execution until the file is uploaded ? I wasn't able to slow it down so that the wait function can come into effect even with a larger file so was unsure of the behavior.

### DIFF
--- a/modelon/impact/client/entities.py
+++ b/modelon/impact/client/entities.py
@@ -209,6 +209,33 @@ class Workspace:
         """
         self._workspace_sal.library_import(self._workspace_id, path_to_lib)
 
+    def upload_result(self, path_to_result, label=None, description=None):
+        """Uploads a '.mat' result file to the workspace.
+
+        Parameters:
+
+            path_to_result --
+                The path for the result file to be imported.
+
+            label --
+                The label of the result file. Default: None.
+
+            description --
+                The description of the result file. Default: None.
+
+        Example::
+
+            workspace.upload_result('C:/A.mat')
+            workspace.upload_result('C:/B.mat', label = "result_for_PID.mat",
+            description = "This is a result file for PID controller")
+        """
+        resp = self._workspace_sal.result_upload(
+            self._workspace_id, path_to_result, label=label, description=description
+        )
+        return operations.ExternalResultUploadOperation(
+            resp["data"]["id"], self._workspace_sal
+        )
+
     def upload_fmu(
         self,
         fmu_path,
@@ -593,10 +620,7 @@ class Workspace:
 
 class _Parameter:
     _JSON_2_PY_TYPE = {
-        "Number": (
-            float,
-            int,
-        ),
+        "Number": (float, int,),
         "String": (str,),
         "Boolean": (bool,),
         "Enumeration": (str,),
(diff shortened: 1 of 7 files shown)


modelon/impact/client/operations.py
```diff
@@ -23,9 +25,23 @@ class Status(Enum):
     DONE = "done"
 
 
-class Operation(ABC):
+@enum.unique
+class UploadStatus(Enum):
```
efredriksson-modelon, line 29: This one is actually a bit more generall than `upload` for example we have this status also for workspace export. So I suggest calling this status 'AsyncOperationStatus'.

modelon/impact/client/operations.py
```diff
@@ -57,6 +73,64 @@ def name(self):
...
 
+    @abstractmethod
+    def wait(self, timeout):
+        """
+        Waits for the operation to finish.
+        """
+        pass
+
+
+class FileOperation(BaseOperation):
```
efredriksson-modelon, line 84: And then this one should be called `AsyncOperation`

tests/impact/client/fixtures.py
```diff
@@ -233,6 +233,87 @@ def import_fmu(sem_ver_check, mock_server_base):
...
+
+@pytest.fixture
+def upload_result_meta(sem_ver_check, mock_server_base):
+    json = {
+        "id": "2f036b9fab6f45c788cc466da327cc78workspace",
+        "createdAt": 1588016253,
+        "name": "result_for_PID",
+        "description": "This is a result file for PID controller",
+        "workspaceId": "workspace",
+    }
```
efredriksson-modelon, line 298: I think the API would return {'data': ...}, this was my point with one of the issues during release testing. Can you dubble check it?

efredriksson-modelon (changes requested):
Good work 👍 Just a few minor comments.

modelon/impact/client/entities.py
```diff
@@ -1996,6 +2053,43 @@ def keys(self):
...
+
+    @property
+    def metadata(self):
+        """External result metadata."""
+        upload_meta = self._workspace_sal.get_uploaded_result_meta(self._result_id)[
+            "data"
+        ]
+        id = upload_meta.get("id")
+        name = upload_meta.get("name")
+        created_at = upload_meta.get("createdAt")
```
efredriksson-modelon, line 2084: If we want to expose the date we should create a datetime object. Here is a stackoverflow on how to convert the isoformat string: https://stackoverflow.com/questions/127803/how-do-i-parse-an-iso-8601-formatted-date. So we would need to install the python-dateutil package to have it work with Python 3.6.

Or we could just remove it for now.

efredriksson-modelon (changes requested):
Just one thing about exposing the date to the user.

modelon/impact/client/entities.py
abhilash-kumar-nair: I have removed it for now. Thanks! 😄

PR #112, opened 2021-09-06
Update case multiple times fix

efredriksson-modelon:
⚠️⚠️ Second commit contains only lint changes, reviewing the commits separately will be easier ⚠️⚠️

This PR fixes the issue where multiple case updates are done and the wrong value for 'consistent' where sent to the server, resulting in a server error. With this PR this code is possible to run with no errors:
```
c = Client()
w = client.create_workspace('test_multiple_updates')
m = w.get_model('Modelica.Blocks.Examples.PID_Controller')
dynamic = w.get_custom_function('dynamic')
exp_def = m.new_experiment_definition(dynamic)
exp = w.create_experiment(exp_def).execute().wait()
case = exp.get_case('case_1')
case.input.parametrization = {'inertia1.J': 5}
case.update()
case.input.parametrization = {'inertia1.J': 6}
case.update()
```

### DIFF
--- a/modelon/impact/client/sal/service.py
+++ b/modelon/impact/client/sal/service.py
@@ -279,7 +279,7 @@ class ExperimentService:
         url = (
             self._base_uri / f"api/workspaces/{workspace_id}/experiments/{exp_id}"
         ).resolve()
-        return self._http_client.put_json_no_responsoe_body(url, body={"label": label})
+        return self._http_client.put_json_no_response_body(url, body={"label": label})
 
     def execute_status(self, workspace_id, experiment_id):
         url = (
@@ -450,7 +450,7 @@ class HTTPClient:
     def delete_json(self, url, body=None):
         RequestJSON(self._context, "DELETE", url, body).execute()
 
-    def put_json_no_responsoe_body(self, url, body=None):
+    def put_json_no_response_body(self, url, body=None):
         RequestJSON(self._context, "PUT", url, body).execute()
 
     def put_json(self, url, body=None):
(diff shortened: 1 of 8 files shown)


modelon/impact/client/sal/service.py
```diff
@@ -450,9 +450,13 @@ def post_json_no_response_body(self, url, body=None):
     def delete_json(self, url, body=None):
         RequestJSON(self._context, "DELETE", url, body).execute()
 
-    def put_json(self, url, body=None):
+    def put_json_no_responsoe_body(self, url, body=None):
```
andrewwren-modelon, line 453: typo for response

andrewwren-modelon (commented):
will we standardize on quotes for Black linting from the client and modelon?

efredriksson-modelon:
> 
> 
> will we standardize on quotes for Black linting from the client and modelon?

I am not sure. We have not done it anywhere else so I changed to using "-S" here also and re-ran the lint command.

abhilash-kumar-nair (approved):
Looks good! Just one doubt - Was the swagger for PUT /workspaces/{workspace}/experiments/{experimentId}/cases/{caseId} 
updated recently? I don't see a response body in the swagger?

efredriksson-modelon:
> 
> 
> Looks good! Just one doubt - Was the swagger for PUT /workspaces/{workspace}/experiments/{experimentId}/cases/{caseId}
> updated recently? I don't see a response body in the swagger?

This seems to not be in the documentation yet but PUT methods typically return the updated data in the response body. We can update the API documentation later.

abhilash-kumar-nair:
> This seems to not be in the documentation yet but PUT methods typically return the updated data in the response body. We can update the API documentation later.

Ah ok! Sure! 💯

PR #114, opened 2021-09-07
feat: possible to check if a case is consistent

efredriksson-modelon:
This PR adds the attribute 'consistent' to the case entity. This will help a user know if they should use the result of a case or not.

Example code on a workspace with at least one experiment with at least one case:
```
all_experiments = workspace.get_experiments()
case = all_experiments[0].get_cases()[0]
case.run_info.consistent
True
case.input.parametrization = {'a': 5}
case.update()
case.run_info.consistent
False
```

### DIFF
--- a/modelon/impact/client/entities.py
+++ b/modelon/impact/client/entities.py
@@ -1566,13 +1566,20 @@ class _CaseRunInfo:
     Class containing Case run info.
     """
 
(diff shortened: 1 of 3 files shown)


abhilash-kumar-nair (approved):
Looks correct! 💯

PR #116, opened 2021-09-17
Remove some unused methods

x-oflisback:
Apparently these are no longer desired, @efredriksson-modelon please sanity check the patch.


efredriksson-modelon (approved):
👍

PR #117, opened 2021-09-27
Sync/update case changes

efredriksson-modelon:
This PR make chagnes to the update functionality after discussion:
 * update method renamed to sync
 * sync is called automatically when executing a case(s). There is a flag for turning this of if needed.


abhilash-kumar-nair (approved):
A much nicer/safer workflow! Nice work!🥇

andrewwren-modelon (approved):
good changes

PR #119, opened 2021-10-14
fix: given empty string as API key, do anon login and don't save key

efredriksson-modelon:
This fixes the issue where you don't have any API key configured (remove api.key file) and would run
```
c = Client(url='https://impact.modelon.com/', interactive=True)
```
and then in the prompt for API key just press enter. Then it would not count this as an Anonymous login in Client (but would do so server side), so you don't get any warning. We would also save this empty string to disk for next time as if it was a valid API key.

### DIFF
--- a/tests/impact/client/test_client.py
+++ b/tests/impact/client/test_client.py
@@ -158,7 +158,7 @@ def test_client_login_fail_lets_user_enter_new_key(_, login_fails):
     cred_manager.get_key_from_prompt.assert_called()
 
 
-def test_client_login_api_key_empty_string_anon_login_and_dont_save_key(sem_ver_check):
+def test_empty_api_key_when_login_then_anon_login_and_dont_save_key(sem_ver_check):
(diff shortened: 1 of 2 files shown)


tests/impact/client/test_client.py
```diff
@@ -156,3 +156,17 @@ def test_client_login_fail_lets_user_enter_new_key(_, login_fails):
     )
 
     cred_manager.get_key_from_prompt.assert_called()
+
+
+def test_client_login_api_key_empty_string_anon_login_and_dont_save_key(sem_ver_check):
```
modelon-magnus, line 161: I would prefer if we could use BDD style acceptance criteria (given, when, then), e.g. "test_given_empty_login_key_when_login_anonymously_then_key_should_not_be_saved"

https://en.wikipedia.org/wiki/Behavior-driven_development

modelon-magnus (commented):
Nice changes! Just have a small comment about naming tests.

PR #120, opened 2021-10-28
Jupyterhub authorization

efredriksson-modelon:
This PR adds support to the client to access an Impact server running in a Jupyterhub. It will be automatically detected if the server the user is pointing against is a Jupyterhub and if that case we will authorize against. This authorization is like how we handle API keys in Impact. Once authorized a new URL and context exists to start the regular Impact client business.

### DIFF
--- a/.releaserc
+++ b/.releaserc
@@ -1,20 +1,18 @@
 {
-  "release": {
-    "branches":[
-      {name: 'master'},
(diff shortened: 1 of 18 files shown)


andrewwren-modelon (approved):
liked the move to the dev user as part of the change

PR #121, opened 2021-11-03
fix: imports that works for Python 3.6

efredriksson-modelon:
This resolves the import issue of the client using Python 3.6.


andrewwren-modelon (approved):
:+1:

hernrup-modelon (approved):
This makes sense and looks good.

PR #122, opened 2021-11-18
Experiment user data

x-oflisback:
This PR adds support for attaching a custom user_data dictionary object to experiments.

It can be added like this:

```
experiment = workspace.create_experiment(experiment_definition, user_data={"customProp": customValue})
```

or like this:

```
experiment = workspace.execute(experiment_definition, user_data={"customProp": customValue}).wait()
```

and then later retrieved:

```
user_data = experiment.metadata.user_data
```

The test additions are probably the main pain-point in this PR, suggestions for improvements are most welcome :pray:

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ make lint
 
 ## Build
 
-Building chimp is done by running
+Building modelon-impact-client is done by running
 
 ```
 make wheel
--- a/modelon/impact/client/entities.py
+++ b/modelon/impact/client/entities.py
@@ -539,7 +539,7 @@ class Workspace:
             resp,
         )
 
-    def create_experiment(self, definition):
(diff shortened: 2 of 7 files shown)


efredriksson-modelon (changes requested):
I think the implementation looks good! 🥇 

Just a few comments on the tests. It might be a good idea to split the tests `test_create_experiment_with_user_data` and `test_execute_with_user_data` because they test both that you can create/execute with user data (and we should verify that the service recives the correct data) and that when you fetch the experiment from server the entity is setup correctly. For fetching the experiment it would be good to have
```
test_get_experiment_no_user_data(self, workspace) # Rename of existing test_get_experiment
test_get_experiment_with_user_data(self, workspace)
```

efredriksson-modelon (approved):
👍

PR #123, opened 2021-11-24
feat: login again if there is an authentication error

efredriksson-modelon:
This PR solves so you will login again if your JWT expires. This is particularly important for long running simulations that run over the night. The implementation is done by a decorator that is applied to all methods of the service classes. This decorator will look for a HTTP 401 failure, re-login, and retry.

### How to test:
```
 c = Client(url='...', interactive=True)
print(c.get_workspaces())
c._sal._http_client._context.session.cookies.clear() # Ugly hack to ensure next request will fail with a 401
print(c.get_workspaces())
```
Previously this code would fail but now it should recover.

### DIFF
--- a/modelon/impact/client/client.py
+++ b/modelon/impact/client/client.py
@@ -77,8 +77,15 @@ class Client:
                 modelon.impact.client.credential_manager.CredentialManager()
             )
 
+        def credential_resolver():
+            return self._authenticate_against_api(
+                interactive=False, anonymous_login_warning=False
+            )
+
         self._uri = modelon.impact.client.sal.service.URI(url)
-        self._sal = modelon.impact.client.sal.service.Service(self._uri, context)
+        self._sal = modelon.impact.client.sal.service.Service(
+            self._uri, context, credential_resolver
+        )
         self._credentials = credential_manager
         self._api_key = None
 
@@ -88,7 +95,9 @@ class Client:
             self._uri, context = modelon.impact.client.jupyterhub.authorize(
                 self._uri, interactive, context, jupyterhub_credential_manager,
             )
-            self._sal = modelon.impact.client.sal.service.Service(self._uri, context)
+            self._sal = modelon.impact.client.sal.service.Service(
+                self._uri, context, credential_resolver
+            )
             self._validate_compatible_api_version()
 
         try:
@@ -121,12 +130,12 @@ class Client:
                 f"that supports version '{version}' of the HTTP REST API."
             )
 
-    def _authenticate_against_api(self, interactive):
+    def _authenticate_against_api(self, interactive, anonymous_login_warning=True):
 
         if not self._api_key:
             self._api_key = self._credentials.get_key(interactive=interactive)
 
-        if not self._api_key:
+        if not self._api_key and anonymous_login_warning:
             logger.warning(
                 "No Modelon Impact API key could be found, "
                 "will log in as anonymous user. Permissions may be limited"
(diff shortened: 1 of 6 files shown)


x-oflisback (commented):
This looks like a real nice decorator implementation to me :rocket:

modelon/impact/client/client.py
```diff
@@ -77,8 +77,15 @@ def __init__(
                 modelon.impact.client.credential_manager.CredentialManager()
             )
 
+        def credential_resolver():
```
andrewwren-modelon, line 80: I am unsure about the readability of this closure and how that mixes in within the need for the complex setattr decorator pattern - is there something that can be done to make simpler to reason about?

modelon/impact/client/sal/service.py
```diff
@@ -24,6 +32,36 @@ def __init__(self, uri, context=None, check_return=True):
...
+                        raise
+
+                    credential_resolver()
+                    return func(*args, **kwargs)
+
+            return wrapped
+
+        self.workspace = _decorate_all_methods(
+            self.workspace, retry_with_login_decorator
+        )
```
modelon-magnus, line 54: I'm not fond of reassigning existing variables. How about having a factory-function that determines if you should decorate or not? E.g. some thing like:
self.workspace = _apply_retry_if_needed(WorkspaceService(...))
In one case it would include the credentials_resolver and decorate the methods, in the other case it would just pass the object.

modelon/impact/client/client.py
efredriksson-modelon: The only alternative I could think of would be to have a decorator lik `retry_with_login_decorator` but have it get access to the `credential_resolver` through `self` and then add this decorator to every method of all service classes.

This solution would be more explicit. The downside is that the credential_resolver functions would be passed around everywhere, and you need to remember to add the decorator to all methods.

modelon-magnus (changes requested):
Good approach with decorator but we shouldn't reassign any variables.

modelon/impact/client/sal/service.py
efredriksson-modelon: That sounds like a good idea. I will try it out.

modelon/impact/client/client.py
andrewwren-modelon: could move all the of the _authenticate_against_api functionality into the service class and mirror the same style as modelon.impact.client.jupyterhub.authorize - in addition the call to api_login which lives in the service class is part of the instance method func wrapper being passed around.

modelon/impact/client/sal/service.py
efredriksson-modelon: I ended up going in another direction. When trying to sort out so there is less implicit dependency on Client workflows in the SAL i realized that we should not add the re-try logic until we have the credentials/API key we want to it with. This leads to the re-try being added later as a separate call to the SAL where everything is much more explicit (there is no callback that in itself would call the SAL...).

modelon/impact/client/client.py
efredriksson-modelon: I belived I have simplified it with less implicit dependencies. Please have a look.

efredriksson-modelon:
@andrewwren-modelon @modelon-magnus Thanks for the reviews! I have refactored it into something I believe is simpler 😄, please have a look.

andrewwren-modelon (approved):
:+1: Thank you for the revisit to the logic

modelon-magnus (approved):
Still not fond of the reassigning of variables but I guess we need the functionality.

PR #124, opened 2021-12-02
Check license after login

efredriksson-modelon:
This PR uses the new version of the API where you can get the license of current user. If no license an exception is raised informing the user. I am very open to improvements to the error message; this was the best I could think of.

### DIFF
--- a/modelon/impact/client/client.py
+++ b/modelon/impact/client/client.py
@@ -56,7 +56,7 @@ class Client:
         client = Client(url=impact_url, interactive=True)
     """
 
-    _SUPPORTED_VERSION_RANGE = ">=1.18.0,<2.0.0"
+    _SUPPORTED_VERSION_RANGE = ">=1.21.0,<2.0.0"
 
     def __init__(
         self,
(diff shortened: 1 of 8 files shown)


modelon/impact/client/client.py
```diff
@@ -105,6 +105,13 @@ def __init__(
 
         self._sal.add_login_retry_with(api_key)
 
+        resp = self._sal.users.get_me()
+        if 'license' not in resp['data']:
+            raise exceptions.NoAssignedLicenseError(
+                "No assigned license during login. Either out of floating "
```
x-oflisback, line 111: Maybe "sharable license" is the established term? At least that's what's used in the ops manual

x-oflisback (approved):
Nice, did not look too close at the test changes.

modelon/impact/client/client.py
efredriksson-modelon: We might decide to change the naming in the ops manual. But we should at least change this error message to use: "<floating|sharable> Deployment Add-on licenses" to be more precise.

efredriksson-modelon: After discussion we went with floating. I also have made a commit where the type of the license is more precise.

andrewwren-modelon (approved):
:+1: lgtm

PR #130, opened 2022-01-21
Api key docs

abhilash-kumar-nair:
This PR(https://modelonproducts.atlassian.net/browse/WAMS-9533) adds docs for JupyterHub auth

### DIFF
--- a/docs/source/setup.rst
+++ b/docs/source/setup.rst
@@ -4,9 +4,39 @@ Setup
 This page describes the procedure to set up API keys to authenticate your login and
 various ways to use the generated API key to log in to the client.
 
+Authentication with JupyterHub
+******************************
+
+This section is ONLY applicable for users running Modelon Impact in a JupyterHub environment,
+otherwise skip to the section :ref:`below<Authentication with API keys>`.
+To use the client on JupyterHub, the users needs to first authenticate with a token
+for Jupyterhub. To generate the key, go to the JupyterHub token manager 
+(located at <JupyterHub url>/token).You can get a secret token by clicking the 
+"Request new API token" button. These are tokens with full access to the JupyterHub API. 
+Anything you can do with JupyterHub can be done with these tokens.
+
+*Remember that this is the only time you will see the secret. Make sure to store it
+safely.*
+
+If the token is lost, you can always generate a new one by clicking the "Request new API token"
+button.
+While initializing the client, you will be asked to enter the JupyterHub API token in a prompt.
+
+.. code-block:: pycon
+
+    >>>> from modelon.impact.client import Client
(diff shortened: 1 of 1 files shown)


docs/source/setup.rst
```diff
@@ -4,9 +4,38 @@ Setup
...
+safely.*
+
+If the token is lost, you can always generate a new one by clicking the "Request new API token"
+button.
+While initializing the client, you will be asked to enter the token key in a prompt.
+
+.. code-block:: pycon
+
+    >>>> from modelon.impact.client import Client
+    >>>> client = Client(url=impact_url)
```
efredriksson-modelon, line 27: we should have `interactive=True` as you will only get a prompt if that is the case.

efredriksson-modelon: Maybe we should have `url=<JupyterHub url>` to be consistent with the rest of the text?

docs/source/setup.rst
```diff
@@ -4,9 +4,38 @@ Setup
...
+(located at <JupyterHub url>/token).You can get a secret token by clicking the 
+"Request new API token" button. These are tokens with full access to the JupyterHub API. 
+Anything you can do with JupyterHub can be done with these tokens.
+
+*Remember that this is the only time you will see the secret. Make sure to store it
+safely.*
+
+If the token is lost, you can always generate a new one by clicking the "Request new API token"
+button.
+While initializing the client, you will be asked to enter the token key in a prompt.
```
efredriksson-modelon, line 22: token key -> JupyterHub API token

docs/source/setup.rst
```diff
@@ -4,9 +4,38 @@ Setup
 This page describes the procedure to set up API keys to authenticate your login and
 various ways to use the generated API key to log in to the client.
 
+Authentication with JupyterHub
+******************************
+
+This section is ONLY applicable for users running Modelon Impact in a JupyterHub environment.
```
efredriksson-modelon, line 10: Maybe we could add something like:
, otherwise skip to :ref:\`below\<Authentication with API keys\>\`.

efredriksson-modelon (changes requested):
Nice with clear instructions on how to do this 👍

Added a few suggestions for clarification.

docs/source/setup.rst
abhilash-kumar-nair: The prompt doesn't require interactive flag set to true to show up?

efredriksson-modelon: You should only get the prompt if `interactive=True`, that is the purpose of the interactive flag.

efredriksson-modelon (approved):
👍

PR #139, opened 2022-02-16
feat: remove workspace locks

x-oflisback:
This PR aims to remove workspace lock/unlock functionality and the corresponding tests.

As a consequence this could mean that the new client API release would require Impact API version 2.0.0 or higher where locks are removed.

However since in practice the python client is only changed to use a subset of the Impact API (i.e. the current Impact API minus locks) it is compatible with older impact versions. So a suggestion is to not update the supported version range to exclude older versions.

### DIFF
--- a/modelon/impact/client/client.py
+++ b/modelon/impact/client/client.py
@@ -56,7 +56,7 @@ class Client:
         client = Client(url=impact_url, interactive=True)
     """
 
-    _SUPPORTED_VERSION_RANGE = ">=1.21.0,<3.0.0"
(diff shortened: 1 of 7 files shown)


efredriksson-modelon (changes requested):
I think the changes looks good 👍 

I think your reasoning about supported versions is mostly correct. However, the version of the API where locks are removed will be 3.0.0 so we should change in this PR to also allow API versions of 3.x.

efredriksson-modelon (approved):
👍

PR #151, opened 2022-06-10
fix: compability for accessing status enums for users

efredriksson-modelon:
This is a fix for a breaking change that was missed due to a rushed release where the integration tests where not updated beforehand.


abhilash-kumar-nair (approved):
Thanks for doing this! The changes look good! :)

PR #153, opened 2022-08-01
fix: added support for new endpoints

abhilash-kumar-nair:
**Jira** - https://modelonproducts.atlassian.net/browse/WAMS-10760

### DIFF
--- a/modelon/impact/client/entities/project.py
+++ b/modelon/impact/client/entities/project.py
@@ -1,0 +1,252 @@
+#
+# Copyright (c) 2022 Modelon AB
+#
+import enum
+from pathlib import Path
+from typing import List, Dict
+from modelon.impact.client.options import (
+    CompilerOptions,
+    RuntimeOptions,
+    SimulationOptions,
+    SolverOptions,
+)
+from modelon.impact.client.sal.project import ProjectService
+
+
+@enum.unique
+class ContentType(enum.Enum):
+    """Supported content types in a project."""
+
+    MODELICA = 'MODELICA'
+    VIEWS = 'VIEWS'
+    FAVOURITES = 'FAVOURITES'
+    CUSTOM_FUNCTIONS = 'CUSTOM_FUNCTIONS'
+    REFERENCE_RESULTS = 'REFERENCE_RESULTS'
+    GENERIC = 'GENERIC'
+
+
+class ProjectContent:
+    """Content entry in a project."""
+
+    def __init__(
+        self, content: Dict[str, str], project_id: str, project_sal: ProjectService
+    ):
+        self._content = content
+        self._project_id = project_id
+        self._project_sal = project_sal
+
+    def __repr__(self):
+        return f"Project content with id '{self.id}'"
+
+    def delete(self):
+        """Deletes a project content.
+
+        Example::
+
+            content.delete()
+        """
+        self._project_sal.project_content_delete(self._project_id, self.id)
+
+    def __eq__(self, obj):
+        return isinstance(obj, ProjectContent) and obj.id == self.id
+
+    @property
+    def relpath(self):
+        """
+        Relative path in the project. Can be file (e.g., SomeLib.mo) or folder
+        """
+        return self._content.get('relpath')
+
+    @property
+    def content_type(self) -> ContentType:
+        """Type of content"""
+        return ContentType(self._content.get('contentType'))
+
+    @property
+    def id(self):
+        """Content ID."""
+        return self._content.get('id')
+
+    @property
+    def name(self):
+        """Modelica library name"""
+        return self._content.get('name')
+
+    @property
+    def default_disabled(self):
+        return self._content.get('defaultDisabled')
+
+
+class ProjectDependency:
+    """Dependency entry for a project"""
+
+    def __init__(self, data):
+        self._data = data
+
+    @property
+    def name(self):
+        """The name of the project dependency"""
+        return self._data.get('name')
+
+    @property
+    def version_specifier(self):
+        """Version specifier"""
+        return self._data.get('versionSpecifier')
+
+
+class ProjectExecutionOptions:
+    def __init__(self, data):
+        self._data = data
+
+    @property
+    def custom_function(self):
+        return self._data.get('customFunction')
+
+    @property
+    def compiler_options(self):
+        return CompilerOptions(self._data.get("compiler"), self.custom_function)
+
+    @property
+    def runtime_options(self):
+        return RuntimeOptions(self._data.get("runtime"), self.custom_function)
+
+    @property
+    def simulation_options(self):
+        return SimulationOptions(self._data.get("simulation"), self.custom_function)
+
+    @property
+    def solver_options(self):
+        return SolverOptions(self._data.get("solver"), self.custom_function)
+
+
+class ProjectDefinition:
+    """
+    Impact project definition.
+    """
+
+    def __init__(self, data):
+        self._data = data
+
+    @property
+    def name(self):
+        return self._data.get("name")
+
+    @property
+    def version(self):
+        return self._data.get("version")
+
+    @property
+    def format(self):
+        return self._data.get("format")
+
+    @property
+    def dependencies(self):
+        dependencies = self._data.get('dependencies', [])
+        return [ProjectDependency(dependency) for dependency in dependencies]
+
+    @property
+    def content(self):
+        return self._data.get('content', [])
+
+    @property
+    def execution_options(self):
+        execution_options = self._data.get('executionOptions', [])
+        return [
+            ProjectExecutionOptions(execution_option)
+            for execution_option in execution_options
+        ]
+
+
+class Project:
+    """
+    Class containing Project functionalities.
+    """
+
+    def __init__(
+        self,
+        project_id: str,
+        project_definition: ProjectDefinition,
+        project_service: ProjectService,
+    ):
+        self._project_id = project_id
+        self._project_definition = project_definition
+        self._project_sal = project_service
+
+    def __repr__(self):
+        return f"Project with id '{self._project_id}'"
+
+    def __eq__(self, obj):
+        return isinstance(obj, Project) and obj._project_id == self._project_id
+
+    @property
+    def id(self) -> str:
+        """Project id"""
+        return self._project_id
+
+    def delete(self):
+        """Deletes a project.
+
+        Example::
+
+            project.delete()
+        """
+        self._project_sal.project_delete(self._project_id)
+
+    def _get_project_content(self, content):
+        return ProjectContent(content, self._project_id, self._project_sal)
+
+    def get_contents(self) -> List[ProjectContent]:
+        """Get project contents.
+
+        Example::
+
+            project.get_contents()
+        """
+        return [
+            self._get_project_content(content)
+            for content in self._project_definition.content
+        ]
+
+    def upload_content(
+        self, path_to_content: str, content_type: ContentType
+    ) -> ProjectContent:
+        """Upload content to a project.
+
+        Parameters:
+
+            path_to_content --
+                The path for the content to be imported.
+
+            content_type --
+                The type of the imported content.
+
+        Example::
+            from modelon.impact.client import ContentType
+
+            project.upload_content('/home/test.mo', ContentType.MODELICA)
+        """
+        resp = self._project_sal.project_content_upload(
+            path_to_content, self._project_id, content_type.value
+        )
+        return self._get_project_content(resp)
+
+    def upload_modelica_library(self, path_to_lib: str):
+        """Uploads/adds a non-encrypted modelica library or a modelica model to the project.
+
+        Parameters:
+
+            path_to_lib --
+                The path for the library to be imported. Only '.mo' or '.zip' file
+                extension are supported for upload.
+
+        Example::
+
+            project.upload_model_library('C:/A.mo')
+            project.upload_model_library('C:/B.zip')
+        """
+        if Path(path_to_lib).suffix.lower() not in ['.mo', '.zip']:
+            raise ValueError(
+                "Only '.mo' or '.zip' file extension are supported for uploading "
+                "Modelica content into project."
+            )
+        return self.upload_content(path_to_lib, ContentType.MODELICA)
--- a/modelon/impact/client/entities/workspace.py
+++ b/modelon/impact/client/entities/workspace.py
@@ -1,4 +1,3 @@
-from pathlib import Path
 import logging
 import os
 from typing import Any, List, Dict, Optional, Union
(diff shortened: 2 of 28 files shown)


modelon/impact/client/entities/workspace.py
```diff
@@ -17,13 +20,63 @@
...
+    default_project_id: str
+
+    projects: List[ProjectEntry]
+
+    dependencies: List[ProjectEntry]
+
+    @classmethod
+    def from_dict(cls, data):
+        projects = data.get('projects', [])
+        dependencies = data.get('dependencies', [])
```
efredriksson-modelon, line 67: I think it would be good to let the workspace definition keep the `data` as a private attribute (probably easier to do this without having it as a data class) so that the data is just passed on when sharing or in other way then serializing and passing on the data.

modelon/impact/client/entities/workspace.py
```diff
@@ -34,22 +87,30 @@ class Workspace:
...
 
     def __repr__(self):
         return f"Workspace with id '{self._workspace_id}'"
 
     def __eq__(self, obj):
-        return isinstance(obj, Workspace) and obj._workspace_id == self._workspace_id
+        return (
+            isinstance(obj, Workspace)
+            and obj._workspace_id == self._workspace_id
+            and obj._workspace_definition == self._workspace_definition
```
efredriksson-modelon, line 112: Should just check the ID here.

modelon/impact/client/entities/project.py
```diff
@@ -0,0 +1,245 @@
...
+            path_to_lib --
+                The path for the library to be imported. Only '.mo' or '.zip' file
+                extension are supported for upload.
+
+        Example::
+
+            project.upload_model_library('C:/A.mo')
+            project.upload_model_library('C:/B.zip')
+        """
+        if Path(path_to_lib).suffix.lower() == '.mol':
```
efredriksson-modelon, line 241: Should this not be `not in ['.mo', '.zip']` rather than `== '.mol'` so the check is what the error bellow says.

modelon/impact/client/entities/workspace.py
```diff
@@ -132,6 +193,26 @@ def upload_model_library(self, path_to_lib: str):
...
+        Parameters:
+
+            path_to_lib --
+                The path for the library to be imported.
+
+        Example::
+
+            workspace.upload_model_library('C:/B.mol')
+        """
+        if Path(path_to_lib).suffix.lower() in ['.mo', 'zip']:
```
efredriksson-modelon, line 214: `!= '.mol'`?

modelon/impact/client/entities/workspace.py
```diff
@@ -132,6 +193,26 @@ def upload_model_library(self, path_to_lib: str):
...
+            raise ValueError("Only '.mol' files are supported for upload.")
```
efredriksson-modelon, line 215: '...for upload as a project dependency' to make the error message more clear.

modelon/impact/client/entities/workspace.py
```diff
@@ -305,27 +386,30 @@ def download(self, options: Dict[str, Any], path: str) -> str:
             f.write(data)
         return ws_path
 
-    def clone(self) -> 'Workspace':
-        """Clones the workspace.
-        Returns a clone Workspace class object.
+    # TODO: Cloning workspace is not implemented on feature branch
+    # def clone(self) -> 'Workspace':
```
efredriksson-modelon, line 390: Would be better to keep this uncommented? This seems a little messy and we probably want the clone method to exist so we should keep it. If we remove methods we will have to do a BREAKING CHANGE commit to get correct versioning.

modelon/impact/client/entities/workspace.py
```diff
@@ -545,3 +629,117 @@ class object.
             self._model_exe_sal,
             self._exp_sal,
         )
+
+    def _refresh(self):
```
efredriksson-modelon, line 633: I don't think we should have these refresh methods that will result in mutation of the data depending on things happening in other places instead of having an explicit update made by the user.

modelon/impact/client/entities/workspace.py
```diff
@@ -545,3 +629,117 @@ class object.
...
+            Project(
+                item["id"],
+                ProjectDefinition.from_dict(item['definition']),
+                self._project_sal,
+            )
+            for item in resp["data"]["items"]
+        ]
+        return projects
+
+    def get_dependencies(self, include_disabled: bool = True):
```
efredriksson-modelon, line 662: Skip `include_disabled` arg, the user can handle this themself if needed. If we find that this is an important feature we should probably implement it into the REST end-point so the client can be stupid.

modelon/impact/client/entities/workspace.py
```diff
@@ -545,3 +629,117 @@ class object.
...
+            project
+            for project in projects
+            if project.id == self._workspace_definition.default_project_id
+        ]
+        if not project:
+            logger.info(
+                f'No default project exists for the workspace {self._workspace_id}!'
+                'Creating a new project and setting it as default project.'
+            )
+            return self.create_project(DEFAULT_PROJECT_NAME)
```
efredriksson-modelon, line 743: I don't think this is what you want. This method should just be something like:
```
if not self.definition.default_project_id:
   raise ValueError('No default project exists for the workspace...')

resp = self._project_sal.project_get(self.definition.default_project_id)
return ...
```

modelon/impact/client/sal/workspace.py
```diff
@@ -120,9 +132,10 @@ def workspace_download(self, workspace_id: str, options: Dict[str, Any]):
         ).resolve()
         return self._http_client.get_zip(url)
 
-    def workspace_clone(self, workspace_id: str):
-        url = (self._base_uri / f"api/workspaces/{workspace_id}/clone").resolve()
-        return self._http_client.post_json(url)
+    # TODO: Cloning workspace is not implemented on feature branch
```
efredriksson-modelon, line 135: same comment about keeping this for now.

modelon/impact/client/entities/workspace.py
```diff
@@ -132,6 +193,26 @@ def upload_model_library(self, path_to_lib: str):
             workspace.upload_model_library('C:/A.mo')
             workspace.upload_model_library('C:/B.mol')
         """
+        logger.warning(
+            "The 'upload_model_library' is deprecated and will be removed in a future"
+            " release. Use 'upload_dependency' method instead!"
+        )
+        self._workspace_sal.library_import(self._workspace_id, path_to_lib)
```
efredriksson-modelon, line 200: It will probably be more sound to just remove this method and end-point because it will not behave as you expect. Better to get users to use the new stuff. So we should have clean commit with BREAKING CHANGE and that removes this method, then we get correct versioning and users can check for breaking changes in the release notes.

efredriksson-modelon (changes requested):
Very nice getting all this done during the summer! 🍉

abhilash-kumar-nair:
@efredriksson-modelon , Thanks for the suggestions. I have now made the suggested fixes.

efredriksson-modelon:
@abhilash-kumar-nair Can you ping me when this PR is rebased? Right now it is very hard to review it.

tests/impact/client/entities/test_project.py
```diff
@@ -0,0 +1,72 @@
...
+from modelon.impact.client import ContentType
+from tests.files.paths import SINGLE_FILE_LIBRARY_PATH
+
+
+class TestProject:
+    def test_get_project_contents(self, project):
+        contents = project.entity.get_contents()
+        assert contents == [
+            create_project_content_entity(
+                project_id="bf1e2f2a2fd55dcfd844bc1f252528f707254425",
```
efredriksson-modelon, line 15: It seems like this magic ID is pasted in a lot of places. It would be nice to only have this string defined in one single place to make the test code more maintainable.

tests/impact/client/entities/test_project.py
```diff
@@ -0,0 +1,72 @@
...
+    def test_get_project_contents(self, project):
+        project_service = project.service
+        content = create_project_content_entity(
+            project_id="bf1e2f2a2fd55dcfd844bc1f252528f707254425",
+            project_service=project.service,
+        )
+        content.delete()
+        project_service.project_content_delete.assert_called_with(
+            "bf1e2f2a2fd55dcfd844bc1f252528f707254425",
+            "81ac23172d7a479db85126691e090b34",
```
efredriksson-modelon, line 29: same with this magic ID for a content entry. Repeated in many places.

tests/impact/client/entities/test_project.py
```diff
@@ -0,0 +1,72 @@
...
+)
+from modelon.impact.client import ContentType
+from tests.files.paths import SINGLE_FILE_LIBRARY_PATH
+
+
+class TestProject:
+    def test_get_project_contents(self, project):
+        contents = project.entity.get_contents()
+        assert contents == [
+            create_project_content_entity(
```
efredriksson-modelon, line 14: Instead of having asserts like these where the expected content class is created by some logic somewhere else it would probably be clearer to make asserts on the fields, this will make it easier to read the code and see what is being tested. Like:
```
assert len(contents) == 1
assert contens[0].id == "..."
```

tests/impact/client/entities/test_project.py
```diff
@@ -0,0 +1,72 @@
...
+            project_id="bf1e2f2a2fd55dcfd844bc1f252528f707254425",
+            content={
+                "id": "f727f04210b94a0fac81f17f83b869e6",
+                "relpath": "test.mo",
+                "contentType": "MODELICA",
+                "name": "test",
+                "defaultDisabled": False,
+            },
+            project_service=project.service,
+        )
```
efredriksson-modelon, line 57: Same here:
```
assert content.id == "..."
assert content.relpath == "test.mo"
...
```

tests/impact/client/entities/test_workspace.py
```diff
@@ -124,3 +124,66 @@ def test_create_experiment_with_user_data(self, workspace):
...
+        assert projects == [
+            create_project_entity(
+                project_id="bf1e2f2a2fd55dcfd844bc1f252528f707254425",
+                project_name="NewProject",
+            )
+        ]
+
+    def test_get_all_dependencies(self, workspace):
+        dependencies = workspace.entity.get_dependencies()
+        assert dependencies == [
```
efredriksson-modelon, line 143: same with this assert. It is going to be very hard to understand what is wrong if this assert fails looking at the code.

tests/impact/client/helpers.py
```diff
@@ -15,17 +20,57 @@
...
     custom_function_service=None,
+    project_service=None,
 ):
+    if not definition:
+        definition = {
+            "name": name,
+            "format": "1.0",
+            "description": "",
+            "createdBy": "local-installation-user-id",
+            "createdAt": 1659072911361,
```
efredriksson-modelon, line 36: Is this correct? Should be a string I think

tests/impact/client/fixtures.py
```diff
@@ -22,6 +23,42 @@
 MockedServer = collections.namedtuple('MockedServer', ['url', 'context', 'adapter'])
 ExperimentMock = collections.namedtuple('ExperimentMock', ['entity', 'service'])
 WorkspaceMock = collections.namedtuple('WorkspaceMock', ['entity', 'service'])
+ProjectMock = collections.namedtuple('ProjectMock', ['entity', 'service'])
+TEST_WORKSPACE_DEFINITION = {
```
efredriksson-modelon, line 27: given how this variable is used it would probably be better to have it like a function:
```
def get_test_workspace_definition(name="newWorkspace"):
  return {
    ...
  }
```
As to not have to creating, copy, and then mutate it.

tests/impact/client/fixtures.py
```diff
@@ -980,22 +1083,149 @@ def workspace():
...
         'data': {'items': [{'id': 'as9f-3df5'}, {'id': 'dd9f-3df5'}]}
     }
     ws_service.workspace_download.return_value = b'\x00\x00\x00\x00'
+    ws_service.workspace_get.return_value = {
+        "definition": {
+            "name": "AwesomeWorkspace",
+            "format": "1.0",
+            "description": "",
+            "createdBy": "local-installation-user-id",
+            "createdAt": 1659072911361,
```
efredriksson-modelon, line 1125: Should be string?

tests/impact/client/fixtures.py
```diff
@@ -980,22 +1083,149 @@ def workspace():
...
+        "projectType": "LOCAL",
+    }
     ws_service.experiment_get.return_value = {'id': 'pid_20090615_134'}
     exp_service.execute_status.return_value = {"status": "done"}
     ws_service.experiments_get.return_value = {
         'data': {'items': [{'id': 'as9f-3df5'}, {'id': 'dd9f-3df5'}]}
     }
     ws_service.workspace_download.return_value = b'\x00\x00\x00\x00'
+    ws_service.workspace_get.return_value = {
+        "definition": {
```
efredriksson-modelon, line 1120: Could the `TEST_WORKSPACE_DEFINITION` data be re-used here?

efredriksson-modelon (changes requested):
Looks very good! 🧇

It would be good to put in some work on the tests to not make them too hard to maintain. We probably want to try collecting information about IDs. Right now it is staring to get very hard to keep track of and there are a lot of magic strings related to IDs in the test code. Maybe we could do something like:
```
class IDs:
  WORKSPACE_PRIMARY = '...'
  WORKSPACE_SECONDARY = '...'
  PROJECT_PRIMARY = '...'
  PROJECT_SECONDARY = '...'

# Test with single WS:
assert workspace.id == IDs.WORKSPACE_PRIMARY

# Test with 2 workspaces:
assert workspaces[0].id == IDs.WORKSPACE_PRIMARY
assert workspaces[1].id == IDs.WORKSPACE_SECONDARY
```
?

tests/impact/client/entities/test_workspace.py
```diff
@@ -1,14 +1,15 @@
 import os
+from re import ASCII
```
efredriksson-modelon, line 2: accidental import

modelon/impact/client/entities/workspace.py
```diff
@@ -17,13 +18,68 @@
...
 from modelon.impact.client.entities.experiment import Experiment
+from modelon.impact.client.entities.project import Project, ProjectDefinition
 
 
 logger = logging.getLogger(__name__)
 
 ExperimentDefinition = Union[
     SimpleModelicaExperimentDefinition, SimpleFMUExperimentDefinition, Dict[str, Any],
 ]
+DEFAULT_PROJECT_NAME = 'NewProject'
```
efredriksson-modelon, line 29: Not used.

tests/impact/client/helpers.py
```diff
@@ -13,19 +18,114 @@
...
+                    "id": IDs.MSL_400_CONTENT_ID,
+                    "name": "MSL",
+                    "version": "4.0.0",
+                },
+                "disabled": False,
+                "disabledContent": [],
+            },
+        ],
+        "id": IDs.WORKSPACE_PRIMARY,
+        "is_clone": False,
```
efredriksson-modelon, line 72: The definition does not contain `id` or `is_clone`

tests/impact/client/helpers.py
```diff
@@ -13,19 +18,114 @@
...
     name,
+    definition=None,
     workspace_service=None,
     model_exe_service=None,
     experiment_service=None,
     custom_function_service=None,
+    project_service=None,
 ):
+    if not definition:
+        definition = {
```
efredriksson-modelon, line 86: `definition = get_test_workspace_definition(name)`?

efredriksson-modelon (changes requested):
Big thanks for going through and fixing all ID refs! 🥇

Just some minor one liner things I found during the final pass and we should be good to go.

abhilash-kumar-nair:
@efredriksson-modelon , Fixed it :)

efredriksson-modelon (approved):
👍

PR #154, opened 2022-08-15
Add instructions for JHub token authentication

thomas-bc:
It took me 5 minutes to dig through the code and find the correct environment variable, so I figured adding it to the docs might save 5 minutes to the next user. Essential piece of documentation for use cases where interactive is not an option (CI etc..).
Thanks!


anna-lindholm:
Hi @thomas-bc,

Thank you very much for your proposed improvement to the documentation! In order to merge the PR, you will need to sign our contributor license agreement (agreement and instructions available at https://github.com/modelon/contributing). Since this is such a small PR, I propose we reject this PR and that we at Modelon add a new one covering this issue. Would that work for you?

Best regards,
Anna

thomas-bc:
Sounds good to me. Thanks.

PR #155, opened 2022-09-02
fix: added instructions for non-interactive JHub token authentication

abhilash-kumar-nair:
Customer request - https://github.com/modelon-community/impact-client-python/pull/154

### DIFF
--- a/modelon/impact/client/jupyterhub/authorize.py
+++ b/modelon/impact/client/jupyterhub/authorize.py
@@ -1,5 +1,5 @@
 import logging
-
+import os
 from modelon.impact.client.credential_manager import CredentialManager
 from modelon.impact.client.jupyterhub import exceptions
 from modelon.impact.client.jupyterhub import sal
@@ -17,9 +17,13 @@ def _get_jupyter_token(credential_manager, interactive, context=None):
 
 
 def authorize(uri, interactive, context=None, credential_manager=None, service=None):
+    # Checking for deprecated environment variable
+    # MODELON_IMPACT_JUPYTERHUB_CLIENT_API_TOKEN for backward compatibility
     credential_manager = credential_manager or CredentialManager(
         file_id="jupyterhub-api.key",
-        env_name="MODELON_IMPACT_JUPYTERHUB_CLIENT_API_TOKEN",
+        env_name="JUPYTERHUB_API_TOKEN"
+        if os.environ.get("JUPYTERHUB_API_TOKEN")
+        else "MODELON_IMPACT_JUPYTERHUB_CLIENT_API_TOKEN",
         interactive_help_text="Enter JupyterHub API token:",
     )
     context = sal.JupyterContext(base=context)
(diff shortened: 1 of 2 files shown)


anna-lindholm:
Looks good, thanks! We might have more feedback on this from other channels which we should also include in the readme, but let's start with this and add a new PR for these.

anna-lindholm (approved):

abhilash-kumar-nair:
@anna-lindholm  Thanks for the review! @iakovn  mentioned that JupyterHub always sets the current token into JUPYTERHUB_API_TOKEN environment variable, hence I have update the code and docs to utilize that environment variable when available instead of asking for a token.

modelon/impact/client/jupyterhub/authorize.py
```diff
@@ -17,9 +17,13 @@ def _get_jupyter_token(credential_manager, interactive, context=None):
...
 
 def authorize(uri, interactive, context=None, credential_manager=None, service=None):
+    # Checking for deprecated environment variable
+    # MODELON_IMPACT_JUPYTERHUB_CLIENT_API_TOKEN for backward compatibility
     credential_manager = credential_manager or CredentialManager(
         file_id="jupyterhub-api.key",
-        env_name="MODELON_IMPACT_JUPYTERHUB_CLIENT_API_TOKEN",
+        env_name="JUPYTERHUB_API_TOKEN"
+        if os.environ.get("JUPYTERHUB_API_TOKEN")
+        else "MODELON_IMPACT_JUPYTERHUB_CLIENT_API_TOKEN",
```
iakovn, line 26: I suggest to deprecate the use of MODELON_IMPACT_JUPYTERHUB_CLIENT_API_TOKEN. We'll be releasing a major version update anyhow and I see no reason for having two different env vars for the same thing.

iakovn (changes requested):
Great to have this info up! 🍬 

Please see the minor comments and then I suggest a bit larger edit both in QuickStart and here:
- QuickStart mentions "<impact-domain>" which seems like a non-standard and as such a confusing term. I suggest to explicitly define that impact domain refers to either  MI cloud or on-prem external URL, e.g., https://impact.modelon.cloud, or a local server, i.e., http://127.0.0.1:8080.  The local server address can be used from a notebook running as a part of cloud or on-prem MI installation.
- I suggest the section in the setup to be renamed into "Authentication with MI cloud or on-prem". Proposed text: 

This part of setup is relevant if you are using MI cloud or on-prem installation and would like to utilize external URL, such as https://impact.modelon.cloud. If you rely on the local connection using, e.g., http://127.0.0.1:8080, then skip to the next section.

MI cloud/on-prem is utilizing JupyterHub as a part of the infrastructure. Therefore authentication when connecting via external URL, such as https://impact.modelon.cloutd requires authenticating with JupyterHub gateway. Connection to the external URL can be done from within a notebook running as a part of MI installation. In that case this will be automatic based on the automatically populated JUPYTERHUB_API_TOKEN environment variable. 

If this client library is used from outside of the environment, e.g., from your desktop, then JupyterHub API token needs to be supplied in JUPYTERHUB_API_TOKEN environment variable or interactively. The token can be obtained from https://impact.modelon.cloud/hub/token page or corresponding page in the on-prem installation.

abhilash-kumar-nair:
Changes suggested have been made in other PRs

PR #157, opened 2022-09-12
fix: adding datetimestamps to run_info

abhilash-kumar-nair:
OTRS ticket - https://otrs.modelon.com/index.pl?Action=AgentTicketZoom;TicketID=40844

### DIFF
--- a/modelon/impact/client/entities/case.py
+++ b/modelon/impact/client/entities/case.py
@@ -1,4 +1,5 @@
 import logging
+from datetime import datetime
 from typing import Any, Dict, Tuple, Optional
 from modelon.impact.client.sal.experiment import ExperimentService
 from modelon.impact.client.sal.model_executable import ModelExecutableService
@@ -23,14 +24,27 @@ def _assert_case_is_complete(status, operation_name="Operation"):
         raise exceptions.OperationFailureError.for_operation(operation_name)
 
 
+def _datetime_from_unix_time(unix_time: Optional[int]):
+    if unix_time:
+        return datetime.fromtimestamp(unix_time / 1e3)
+
+
 class _CaseRunInfo:
     """
     Class containing Case run info.
     """
 
-    def __init__(self, status: CaseStatus, consistent: bool):
+    def __init__(
+        self,
+        status: CaseStatus,
+        consistent: bool,
+        datetime_started: Optional[datetime],
+        datetime_finished: Optional[datetime],
+    ):
         self._status = status
         self._consistent = consistent
+        self._datetime_started = datetime_started
+        self._datetime_finished = datetime_finished
 
     @property
     def status(self) -> CaseStatus:
@@ -42,6 +56,20 @@ class _CaseRunInfo:
         """True if the case has not been synced since it was executed,
         false otherwise."""
         return self._consistent
+
+    @property
+    def started(self):
+        """
+        Case execution start time. Returns None if case execution hasn't started.
+        """
+        return self._datetime_started
+
+    @property
+    def finished(self):
+        """
+        Case execution finish time. Returns None if case execution hasn't finished.
+        """
+        return self._datetime_finished
 
 
 class _CaseAnalysis:
@@ -216,7 +244,11 @@ class Case:
     def run_info(self) -> _CaseRunInfo:
         """Case run information"""
         run_info = self._info["run_info"]
-        return _CaseRunInfo(CaseStatus(run_info["status"]), run_info["consistent"])
+        started = _datetime_from_unix_time(run_info.get("datetime_started"))
+        finished = _datetime_from_unix_time(run_info.get("datetime_finished"))
+        return _CaseRunInfo(
+            CaseStatus(run_info["status"]), run_info["consistent"], started, finished,
+        )
 
     @property
     def input(self) -> _CaseInput:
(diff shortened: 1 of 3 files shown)


modelon/impact/client/entities/case.py
```diff
@@ -28,9 +29,17 @@ class _CaseRunInfo:
     Class containing Case run info.
     """
 
-    def __init__(self, status: CaseStatus, consistent: bool):
+    def __init__(
+        self,
+        status: CaseStatus,
+        consistent: bool,
+        datetime_started: int,
+        datetime_finished: int,
```
efredriksson-modelon, line 37: It would be better if a `datetime` was sent in instead of `int`.

efredriksson-modelon: Also, I think there are cases where `datetime_finished` is None? For example if the case has not finished running. Not sure about `datetime_started` and if it always have to be set.

modelon/impact/client/entities/case.py
```diff
@@ -43,6 +52,16 @@ def consistent(self) -> bool:
         false otherwise."""
         return self._consistent
 
+    @property
+    def datetime_started(self):
```
efredriksson-modelon, line 56: The type of the returned object is `datetime`, just naming this method `started` would be nicer.

modelon/impact/client/entities/case.py
```diff
@@ -43,6 +52,16 @@ def consistent(self) -> bool:
...
+        """Case execution start time."""
+        return datetime.fromtimestamp(self._datetime_started / 1e3)
+
+    @property
+    def datetime_finished(self):
```
efredriksson-modelon, line 61: Same on the name.

efredriksson-modelon (changes requested):
Looks like a nice addition 👍

modelon/impact/client/entities/case.py
abhilash-kumar-nair: Good catch! I looked up the code(see below) and it looks like datetime_started couldn't be none, but datetime_finished could. I have now added checks for both to not get errors and a logger message in case any of them is None.
```
class CaseRunInfo:
    def __init__(
        self,
        status=CaseStatus.NOT_STARTED,
        datetime_started=None,
        datetime_finished=None,
        failed_at=CaseFailedAt.NONE,
        consistent=False,
    ):
        self.status = status
        self.datetime_started = datetime_started or datetime.datetime.utcnow()
        self.datetime_finished = datetime_finished
```

modelon/impact/client/entities/case.py
```diff
@@ -216,7 +239,19 @@ def info(self) -> Dict[str, Any]:
...
+        started = (
+            datetime.fromtimestamp(run_info["datetime_started"] / 1e3)
+            if run_info["datetime_started"]
+            else None
+        )
+        finished = (
+            datetime.fromtimestamp(run_info["datetime_finished"] / 1e3)
+            if run_info["datetime_finished"]
+            else None
+        )
```
efredriksson-modelon, line 251: Maybe this can be factored out like:
`finished = _datetime_from_unix_time(run_info.get("datetime_finished"))`
?

modelon/impact/client/entities/case.py
```diff
@@ -43,6 +52,20 @@ def consistent(self) -> bool:
         false otherwise."""
         return self._consistent
 
+    @property
+    def started(self):
+        """Case execution start time."""
+        if not self._datetime_started:
+            logger.warning("This attribute is deprectated, use 'run_info' instead")
```
efredriksson-modelon, line 59: I don't understand this. Why is this method deprecated only under certain conditions? Also why is this method added if it is directly deprecated?

efredriksson-modelon (changes requested):

modelon/impact/client/entities/case.py
efredriksson-modelon: Ok, I see. I think this should be documented in the doc-string rather than being a warning.

modelon/impact/client/entities/case.py
```diff
@@ -43,6 +57,20 @@ def consistent(self) -> bool:
         false otherwise."""
         return self._consistent
 
+    @property
+    def started(self):
+        """Case execution start time."""
+        if not self._datetime_started:
+            logger.warning("The case execution hasn't started yet!")
```
efredriksson-modelon, line 64: I think it would be much better to document that property is `None` if execution has not started instead of logging a warning.

modelon/impact/client/entities/case.py
```diff
@@ -43,6 +57,20 @@ def consistent(self) -> bool:
...
+        return self._datetime_started
+
+    @property
+    def finished(self):
+        """Case execution finish time."""
+        if not self._datetime_started:
+            logger.warning("The case execution hasn't finished yet!")
```
efredriksson-modelon, line 71: Same, document instead of warning.

efredriksson-modelon (changes requested):

modelon/impact/client/entities/case.py
abhilash-kumar-nair: Fixed it :)

efredriksson-modelon (approved):
👍

PR #159, opened 2022-09-15
New api fixes

efredriksson-modelon:
A few fixes so the client uses the API correctly.


abhilash-kumar-nair (approved):
Thanks for fixing this! The changes/additions look good! 🌞

PR #160, opened 2022-09-26
fix: added convinience methods and arguments for fetching projects

abhilash-kumar-nair:


modelon/impact/client/client.py
```diff
@@ -258,7 +258,7 @@ def get_workspaces(self):
             for item in resp["data"]["items"]
         ]
 
-    def get_project(self, project_id):
+    def get_project(self, project_id: str, vcs_info: bool = True):
```
efredriksson-modelon, line 261: Should default be `True`?

modelon/impact/client/entities/workspace.py
```diff
@@ -157,6 +157,9 @@ def id(self) -> str:
         """Workspace id"""
         return self._workspace_id
 
+    def definition(self):
```
efredriksson-modelon, line 160: property?

efredriksson-modelon (changes requested):
Looks ok from what I can see. 👍 Do we want `vcs_info` `True` by default?

modelon/impact/client/client.py
abhilash-kumar-nair: Yes, we agreed on having this True in the group chat right?

iakovn: From me trying the client out - yes, True is preferable/safer default. This is mostly to give expected result for someone not reading the docs before trying out the code.

iakovn (approved):
Looking good! 👍

PR #162, opened 2022-10-18
fix: concatination of URLs handles starting slash on relative part

efredriksson-modelon:
Some of the relative URLs added in new methods where starting with a slash. This would lead to incorrect concatenated results for some cases. This PR changes the URI class so it does not care about this.


abhilash-kumar-nair (approved):
Good catch! Nice that added tests too. Good to merge 👍

PR #163, opened 2022-12-02
fix: corrected minor typo

abhilash-kumar-nair:


efredriksson-modelon (approved):
👍 

(one minor improvement that could be done is to mention that this typo fix is for the docs in the commit message, so the changelog becomes more readable for users)

PR #165, opened 2022-12-09
Export endpoint

abhilash-kumar-nair:


modelon/impact/client/entities/workspace.py
```diff
@@ -254,7 +255,51 @@ def upload_result(
         )
         return ExternalResultUploadOperation(resp["data"]["id"], self._sal)
 
-    def download(self, options: Dict[str, Any], path: str) -> str:
+    def export(self, options: Dict[str, Any]):
+        """Exports the workspace as a binary compressed archive.
+        Returns an modelon.impact.client.operations.workspace
+        .WorkspaceExportOperation class object.
```
efredriksson-modelon, line 261: Because the export and download end-points are so similar I think it would be good if there was some cross referencing in the docstrings of them. Like 'Similar to 'download' but gives more control for getting the workspace async' and 'Similar to 'export' but does the entire setup and download as in one go'. Not sure about best ways.

modelon/impact/client/entities/workspace.py
```diff
@@ -254,7 +255,51 @@ def upload_result(
...
+                        "_nics_multibody_examples_elementary_doublependulum_20191029",
+                        "modelica_blocks_examples_pid_controller_20191023_151659_f32a30d",
+                    ],
+                    "fmu_ids": [
+                        "_nics_multibody_examples_elementary_doublependulum_20191029_089",
+                        "modelica_blocks_examples_pid_controller_20191023_151659_f32a30d",
+                    ],
+                }
+            }
+            path = workspace.download(options).wait().download_as('/home/workspace.zip')
```
efredriksson-modelon, line 296: download -> export

modelon/impact/client/operations/workspace/exports.py
```diff
@@ -0,0 +1,104 @@
...
+        Returns:
+
+            A BinaryBlob class object.
+        """
+        info = self._info()
+        if info['status'] == AsyncOperationStatus.ERROR.value:
+            raise exceptions.IllegalWorkspaceExport(
+                f"Workspace export failed! Cause: {info['error'].get('message')}"
+            )
+        return BinaryBlob(self._sal.export.export_download(info["data"]["downloadUri"]))
```
efredriksson-modelon, line 87: This means that the download is happening here and not when you call the download_as method. Should be something like:
`BinaryBlob(self._sal.export, info["data"]["downloadUri"])`

modelon/impact/client/operations/workspace/exports.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def cancel(self):
+        raise NotImplementedError('Cancel is not supported for this operation')
+
+    def _info(self):
+        return self._sal.workspace.get_workspace_export_status(self._location)["data"]
+
+    def data(self):
+        """
+        Returns a new Workspace class instance.
```
efredriksson-modelon, line 76: no?

modelon/impact/client/operations/workspace/exports.py
```diff
@@ -0,0 +1,104 @@
+import os
+from modelon.impact.client.sal.service import Service
+from modelon.impact.client.operations.base import AsyncOperation, AsyncOperationStatus
+from modelon.impact.client import exceptions
+
+
+class BinaryBlob:
```
efredriksson-modelon, line 7: BinaryBlob -> Export, would be more clear. It is not representing a binary blob it is representing an URL to something that is exported and ready for download.

tests/impact/client/entities/test_workspace.py
```diff
@@ -46,6 +47,10 @@ def test_download_workspace(self, workspace):
         resp = workspace.entity.download({}, tempfile.gettempdir())
         assert resp == t
 
+    def test_export_workspace(self, workspace):
+        resp = workspace.entity.export({})
+        assert resp == create_workspace_export_operation("79sd8-3n2a4-e3t24")
```
efredriksson-modelon, line 52: 79sd8-3n2a4-e3t24 is not in `IDs`?

efredriksson-modelon (changes requested):
I really like how the API turned out! 🧇

At some point we probably also want to support downloading the files like a stream:
```
with requests.get(url, auth=auth, stream=True) as r:
        with open(path, 'wb') as f:
            for chunk in r.iter_content(chunk_size=8192): 
                f.write(chunk)
```
so we do not have to have the entire file loaded in memory and the write it in one go. But, this is a separate PR at some other time. Seems like it would be a pretty deep change.

modelon/impact/client/operations/workspace/exports.py
```diff
@@ -0,0 +1,104 @@
...
+
+        Returns:
+
+            path --
+                Local path to the downloaded archive.
+
+        Example::
+
+            path = workspace.download(options).wait().download_as('/home/workspace.zip')
+            path = workspace.download(options).wait().download_as('workspace.zip')
```
efredriksson-modelon, line 28: download -> export

abhilash-kumar-nair:
> At some point we probably also want to support downloading the files like a stream:

Thanks @efredriksson-modelon  for the review! I agree about writing data in chunks in future. Could you please take another look?

modelon/impact/client/entities/workspace.py
```diff
@@ -284,24 +338,20 @@ def download(self, options: Dict[str, Any], path: str) -> str:
...
         """
-        data = self._sal.workspace.workspace_download(self._workspace_id, options)
         ws_path = os.path.join(path, self._workspace_id + ".zip")
-
-        os.makedirs(os.path.dirname(path), exist_ok=True)
-        with open(ws_path, "wb") as f:
-            f.write(data)
-        return ws_path
+        ops = self.export(options).wait()
+        return ops.download_as(ws_path) if ops else None
```
efredriksson-modelon, line 354: `ops` can never be `None`?

modelon/impact/client/entities/workspace.py
```diff
@@ -22,7 +23,9 @@
 logger = logging.getLogger(__name__)
 
 ExperimentDefinition = Union[
-    SimpleModelicaExperimentDefinition, SimpleFMUExperimentDefinition, Dict[str, Any],
+    SimpleModelicaExperimentDefinition,
+    SimpleFMUExperimentDefinition,
+    Dict[str, Any],
```
efredriksson-modelon, line 28: Similar changes has been made to master. You are probably aware but merging into dev and rebasing this one would be good.

efredriksson-modelon (changes requested):
Looks good! 👍  Just one minor thing that looked a bit strange.

modelon/impact/client/entities/workspace.py
abhilash-kumar-nair: Ah yes. Sure!

PR #166, opened 2022-12-12
Missing documentation fixes

abhilash-kumar-nair:
This PR fixes the missing "autogenerated" docs for some of the modules in the RTD helpcenter. The sphinx-api-doc cli used to autogenerate the docs needs to be rereun if there are modules being added removed so that the respective rst in the docs/sources could be updated to reflect the structure of the source code. I have now added a convivence make command to "force" generate the regenerate the rst files used to render the docs. This command needs to be run if there are new modules added or existing modules removed.

I intend to merge master to the dev branches with this change. Once this is in. :)


efredriksson-modelon (approved):
Thanks for going the extra mile with updating the packages and adding a make target for the task! 🍉

PR #167, opened 2022-12-16
Workspace conversion

efredriksson-modelon:
### How to test:
You can test it by running:
 * poetry install
 * poetry shell
 * And then execute the Python code:
```

c = Client('<server URL>')
workspace_id = '<workspace ID>'
workspace = c.convert_workspace(workspace_id).wait()
print(workspace.definition.name)
```
on a workspace that should be converted.

### DIFF
--- a/modelon/impact/client/operations/workspace/conversion.py
+++ b/modelon/impact/client/operations/workspace/conversion.py
@@ -1,0 +1,76 @@
+from modelon.impact.client import exceptions
+from modelon.impact.client.entities.workspace import Workspace, WorkspaceDefinition
+from modelon.impact.client.operations.base import AsyncOperation, AsyncOperationStatus
+from modelon.impact.client.sal.service import Service
+
+
+class WorkspaceConversionOperation(AsyncOperation):
+    """
+    An conversion operation class for the modelon.impact.client.entities.workspace.
+    Workspace class.
+    """
+
+    def __init__(self, location: str, service: Service):
+        super().__init__()
+        self._location = location
+        self._sal = service
+
+    def __repr__(self):
+        return f"Workspace conversion operations for id '{self.id}'"
+
+    def __eq__(self, obj):
+        return (
+            isinstance(obj, WorkspaceConversionOperation)
+            and obj._location == self._location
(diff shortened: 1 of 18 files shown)


efredriksson-modelon:
@abhilash-kumar-nair Sure, note that this one is WIP for now and that I have not added tests yet 😃

efredriksson-modelon:
@abhilash-kumar-nair Now it is all ready 😃

modelon/impact/client/operations/workspace/conversion.py
```diff
@@ -0,0 +1,76 @@
+from modelon.impact.client import exceptions
+from modelon.impact.client.entities.workspace import Workspace, WorkspaceDefinition
+from modelon.impact.client.operations.base import AsyncOperation, AsyncOperationStatus
+from modelon.impact.client.sal.service import Service
+
+
+class WorkspaceConversionOperation(AsyncOperation):
+    """
+    An export operation class for the modelon.impact.client.entities.workspace.
```
abhilash-kumar-nair, line 9: export --> conversion

abhilash-kumar-nair (commented):
Really nice to have the conversion workflow implemented with the client. Just a comment/question, how would the user know he needs to convert, would that be a failed GET /workspace_id call that indicated that the user needs to call convert? I was thinking if we had some means to know if a conversion is needed and raise appropriate exceptions for get_workspace() OR automatically do the conversion as a part of the get_workspace() call?

efredriksson-modelon:
> Really nice to have the conversion workflow implemented with the client. Just a comment/question, how would the user know he needs to convert, would that be a failed GET /workspace_id call that indicated that the user needs to call convert? I was thinking if we had some means to know if a conversion is needed and raise appropriate exceptions for get_workspace() OR automatically do the conversion as a part of the get_workspace() call?

You will get
bild

which should make it clear that the workspace needs to be converted. I think having this as a separate step and not something that happens automatically is a feature as the user might not want to convert their workspace.

abhilash-kumar-nair (approved):
Makes sense! 👍

PR #170, opened 2022-12-20
External result fix

efredriksson-modelon:
This PR fixes so that we catch if the external result upload operation status is 'error'. I also fixed lint and formatting errors for the fixture.py file.


abhilash-kumar-nair (approved):
Good catch! Thanks for fixing the linting errors! 👍

PR #172, opened 2023-01-10
Token api key url hint

efredriksson-modelon:
This PR adds hints for the interactive help text during login with URL to where key/token can be generated. This should in particularly help with users not know for sure where the boundary between JH and Impact goes.

login-interactive

### How to test
Run the following code (be sure you do not already have setup token/key for impact.modelon.cloud)
```

client = Client('https://impact.modelon.cloud/', interactive=True)
```

And follow the prompts to where token/key can be generated:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@ RUN pip install -U pip
 
 # update 
 RUN apt-get update && apt-get install -y curl apt-utils bash-completion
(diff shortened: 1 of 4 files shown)


anna-lindholm (approved):
Thanks Emil, this is really a great improvement that I think will help A LOT! Great initiative!

iakovn (approved):
👍

PR #173, opened 2023-01-24
fix: assert compilation is successful before calling download FMU api

abhilash-kumar-nair:
```

client = Client(url="https://jhmi.modelon.com/", interactive=True,)
workspace = client.get_workspace("test")

# Choose analysis type

# Get model
model = workspace.get_model("Test.PID_Controller")
compiler_options = dynamic.get_compiler_options().with_values(c_compiler='gcc')
runtime_options = {'cs_solver': 0}
fmu = model.compile(compiler_options, runtime_options).wait()
# Compilation fails
print(fmu.is_successful())
# Output :- False

# Trying to call the download method on the fmu object
fmu_path = fmu.download('/home/jovyan/impact/workspaces/test')
# Raises OperationFailureError: Compilation failed! See the log for more info!

# Cause of failure
log = fmu.get_log()
log.show()
# Output :-
# Error in flattened model:
#     Index reduction failed: There were unmatched equations and/or variables left after index reduction.
# Error in flattened model:
#     The system is structurally singular. The following variable(s) could not be matched to any equation:
#      der(spring.w_rel)
#      torque.flange.phi

#   The following equation(s) could not be matched to any variable:
#     spring.w_rel = der(spring.phi_rel)
#     0.0 = 0.0

```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@ RUN pip install -U pip
 
 # update 
 RUN apt-get update && apt-get install -y curl apt-utils bash-completion vim
-RUN curl -sL https://deb.nodesource.com/setup_14.x | bash
+RUN curl -sL https://deb.nodesource.com/setup_18.x | bash
 RUN apt-get install -y nodejs
 RUN node -v
 RUN npm -v
--- a/modelon/impact/client/entities/model_executable.py
+++ b/modelon/impact/client/entities/model_executable.py
@@ -238,11 +238,18 @@ class ModelExecutable:
             path --
                 Local path to the downloaded FMU.
 
-        Example::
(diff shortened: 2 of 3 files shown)


anna-lindholm (approved):
Great additions, thanks Abhilash!

efredriksson-modelon (changes requested):
Nice fix! 👍 I don't think the quick start script has to change, in fact now it would behave as expected.

abhilash-kumar-nair:
> Nice fix! 👍 I don't think the quick start script has to change, in fact now it would behave as expected.

Makes sense! Fixed it! :)

efredriksson-modelon (approved):
👍

PR #175, opened 2023-02-21
fix: updated the route for projects endpoint

abhilash-kumar-nair:


efredriksson-modelon (approved):
Thanks for keeping up with needed changes also in the client! :watermelon:

PR #177, opened 2023-03-10
feat: add support to fetch artifact ids

abhilash-kumar-nair:
Jira - https://modelonproducts.atlassian.net/browse/WAMS-11860

### DIFF
--- a/modelon/impact/client/entities/case.py
+++ b/modelon/impact/client/entities/case.py
@@ -1,6 +1,9 @@
 import logging
+import os
+import tempfile
 from datetime import datetime
-from typing import Any, Dict, Tuple, Optional
+from dataclasses import dataclass
+from typing import Any, Dict, Tuple, Optional, List
 from modelon.impact.client.sal.service import Service
 from modelon.impact.client.sal.experiment import ResultFormat
 from modelon.impact.client.operations.case import CaseOperation
@@ -125,6 +128,67 @@ class _CaseAnalysis:
     @simulation_log_level.setter
     def simulation_log_level(self, simulation_log_level: str):
         self._analysis['simulation_log_level'] = simulation_log_level
+
+
+@dataclass
+class CustomArtifact:
+    """CustomArtifact class"""
+
+    def __init__(
+        self,
+        workspace_id: str,
+        experiment_id: str,
+        case_id: str,
+        data: Dict[str, str],
+        exp_sal,
+    ):
+        self._workspace_id = workspace_id
+        self._exp_id = experiment_id
+        self._case_id = case_id
+        self._data = data
+        self._exp_sal = exp_sal
+
+    @property
+    def id(self):
+        """Id of the custom artifact."""
+        return self._data['id']
+
+    @property
+    def download_as(self):
+        """File name for the downloaded artifact."""
+        return self._data['downloadAs']
+
+    def download(self, path: Optional[str] = None):
+        """Downloads a custom artifact.
+        Returns the local path to the downloaded artifact.
+
+        Parameters:
+
+            path --
+                The local path to store the custom artifact. Default: None.
+                If no path is given, custom artifact will be downloaded in a temporary
+                 directory.
+
+        Returns:
+
+            path --
+                Local path to the downloaded custom artifact.
+
+        Example::
+
+            artifact_path = artifact.download()
+            artifact_path = artifact.download('/home/Downloads')
+        """
+        artifact, file_name = self._exp_sal.case_artifact_get(
+            self._workspace_id, self._exp_id, self._case_id, self.id
+        )
+        if path is None:
+            path = os.path.join(tempfile.gettempdir(), "impact-downloads")
+        os.makedirs(path, exist_ok=True)
+        artifact_path = os.path.join(path, file_name)
+        with open(artifact_path, "wb") as f:
+            f.write(artifact)
+        return artifact_path
 
 
 class _CaseMeta:
@@ -465,6 +529,39 @@ class Case:
         )
 
         return result, file_name
+
+    def get_artifacts(self) -> List[CustomArtifact]:
+        """
+        Returns a list of CustomArtifact classes for a finished case.
+
+        Returns:
+
+            custom artifacts --
+                A list of CustomArtifact class objects.
+
+        Raises:
+
+            OperationNotCompleteError if simulation process is in progress.
+            OperationFailureError if simulation process has failed or was cancelled.
+
+        Example::
+
+            custom_artifacts = case.get_artifacts()
+        """
+        assert_successful_operation(self.is_successful(), self._case_id)
+        resp = self._sal.experiment.case_artifacts_meta_get(
+            self._workspace_id, self._exp_id, self._case_id
+        )
+        return [
+            CustomArtifact(
+                self._workspace_id,
+                self.experiment_id,
+                self._case_id,
+                meta,
+                self._sal.experiment,
+            )
+            for meta in resp["data"]["items"]
+        ]
 
     def get_fmu(self):
         """
--- a/modelon/impact/client/entities/case.py
+++ b/modelon/impact/client/entities/case.py
@@ -2,7 +2,6 @@ import logging
 import os
 import tempfile
 from datetime import datetime
-from dataclasses import dataclass
 from typing import Any, Dict, Tuple, Optional, List
 from modelon.impact.client.sal.service import Service
 from modelon.impact.client.sal.experiment import ResultFormat
@@ -130,7 +129,6 @@ class _CaseAnalysis:
         self._analysis['simulation_log_level'] = simulation_log_level
 
 
-@dataclass
 class CustomArtifact:
     """CustomArtifact class"""
 
@@ -139,24 +137,26 @@ class CustomArtifact:
         workspace_id: str,
         experiment_id: str,
         case_id: str,
-        data: Dict[str, str],
+        artifact_id: str,
+        download_as: str,
         exp_sal,
     ):
         self._workspace_id = workspace_id
         self._exp_id = experiment_id
         self._case_id = case_id
-        self._data = data
+        self._artifact_id = artifact_id
+        self._download_as = download_as
         self._exp_sal = exp_sal
 
     @property
     def id(self):
         """Id of the custom artifact."""
-        return self._data['id']
+        return self._artifact_id
 
     @property
     def download_as(self):
         """File name for the downloaded artifact."""
-        return self._data['downloadAs']
+        return self._download_as
 
     def download(self, path: Optional[str] = None):
         """Downloads a custom artifact.
@@ -165,9 +165,9 @@ class CustomArtifact:
         Parameters:
 
             path --
-                The local path to store the custom artifact. Default: None.
-                If no path is given, custom artifact will be downloaded in a temporary
-                 directory.
+                The local path to the directory to store the downloaded custom artifact.
+                Default: None. If no path is given, custom artifact will be downloaded
+                in a temporary directory.
 
         Returns:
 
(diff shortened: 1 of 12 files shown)


tests/impact/client/helpers.py
```diff
@@ -157,11 +157,14 @@ class IDs:
     FMU_SECONDARY = 'test_filter_fmu_id'
     EXPERIMENT_PRIMARY = 'pid_20090615_134'
     EXPERIMENT_SECONDARY = 'filter_20090615_135'
+    CASE_PRIMARY = 'case_1'
```
iakovn, line 160: Can you factor out introduction of constants and corresponding renames across tests into a separate "chore" commit? Mixing that with the new API makes it difficult to separate feature from test improvement.

iakovn (changes requested):
Nice that it gets propagated into client!
See comments.

modelon/impact/client/entities/case.py
```diff
@@ -466,6 +466,12 @@ def get_artifact(self, artifact_id: str) -> Tuple[bytes, str]:
 
         return result, file_name
 
+    def get_artifact_ids(self):
```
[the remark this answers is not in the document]
efredriksson-modelon: I think it would be even better to even have: `get_artifacts() -> list[CustomArtifact]` to future proof it for even more fields. Usage would look something like:
```
>case.get_artifacts()
[CustomArtifact(id='ABCD', download_as='ABCD.mat')]
>case.get_artifacts()[0].download()  # Can be implemented later if we do not want to this right now
```

modelon/impact/client/sal/experiment.py
```diff
@@ -161,3 +161,13 @@ def case_artifact_get(
         ).resolve()
         resp = self._http_client.get_octet_response(url)
         return resp.stream, resp.file_name
+
+    def case_artifact_meta_get(
```
efredriksson-modelon, line 165: Plural on artifact

efredriksson-modelon (changes requested):
I think my main concern is exposing the IDs directly in a list to the user. Would be much better to have an object representing the custom artifact. The existing `get_artifact` method intreface is a bit strange, not sure if we want to let users continue using it or deprecate it for a new method on a CustomArtifact class? You will have to decide on the scope for that.

iakovn:
> I think my main concern is exposing the IDs directly in a list to the user. Would be much better to have an object representing the custom artifact. The existing `get_artifact` method intreface is a bit strange, not sure if we want to let users continue using it or deprecate it for a new method on a CustomArtifact class? You will have to decide on the scope for that.

 @efredriksson-modelon  I agree and we should change `get_artifact(artifact_id: str) → Tuple[bytes, str]` somehow. However I'm not so sure about signature for "download". How about CustomArtifact.write(file) instead?

How shall we go about multisim where artifact ID is the same for all cases. It seems unnecessary to have an extra call for each case just to get the ID without downloading?

efredriksson-modelon:
> > I think my main concern is exposing the IDs directly in a list to the user. Would be much better to have an object representing the custom artifact. The existing `get_artifact` method intreface is a bit strange, not sure if we want to let users continue using it or deprecate it for a new method on a CustomArtifact class? You will have to decide on the scope for that.
> 
> @efredriksson-modelon I agree and we should change `get_artifact(artifact_id: str) → Tuple[bytes, str]` somehow. However I'm not so sure about signature for "download". How about `CustomArtifact.write(stream)` instead?
> 
> How shall we go about multisim where artifact ID is the same for all cases. It seems unnecessary to have an extra call for each case just to get the ID without downloading?

@iakovn Yes, the 'download' I wrote was mostly to signify that it is done on the CustomArtifact object.

I guess the case with a lot of cases where all have the same custom artifacts you would still like the `get_artifact` that exists now to optimize this. I would guess this is not a very common case where you have a lot of cases and you also want to download the same artifact from many of these cases?

iakovn:
> 
> I guess the case with a lot of cases where all have the same custom artifacts you would still like the `get_artifact` that exists now to optimize this. I would guess this is not a very common case where you have a lot of cases and you also want to download the same artifact from many of these cases?

If someone writes a custom function that generates artifacts or writes artifacts directly from the model (via external functions) it'll be exactly the case. One approach could be to allow switching the caseId on the CustomArtifacts object (i.e., introduce `CustomArtifact.with_case(case) -> CustomArtifact`).  What do you think?

tests/impact/client/helpers.py
abhilash-kumar-nair: Ah sorry for that. I will do that before merging.

modelon/impact/client/entities/case.py
```diff
@@ -127,6 +130,67 @@ def simulation_log_level(self, simulation_log_level: str):
         self._analysis['simulation_log_level'] = simulation_log_level
 
 
+@dataclass
```
efredriksson-modelon, line 133: From the implementation of this class does not seem like it is a dataclass?

modelon/impact/client/entities/case.py
```diff
@@ -127,6 +130,67 @@ def simulation_log_level(self, simulation_log_level: str):
...
+        return self._data['downloadAs']
+
+    def download(self, path: Optional[str] = None):
+        """Downloads a custom artifact.
+        Returns the local path to the downloaded artifact.
+
+        Parameters:
+
+            path --
+                The local path to store the custom artifact. Default: None.
```
efredriksson-modelon, line 168: It could be clarified that this local path should is a directory and the file is put in it.

modelon/impact/client/entities/case.py
```diff
@@ -430,23 +494,45 @@ def get_trajectories(self) -> Result:
...
-                artifact stream.
+        Example::
+
+            custom_artifact = case.get_artifact(artifact_id)
+        """
+        assert_successful_operation(self.is_successful(), self._case_id)
+        resp = self._sal.experiment.case_artifacts_meta_get(
+            self._workspace_id, self._exp_id, self._case_id
+        )
+        meta = [meta for meta in resp["data"]["items"] if meta["id"] == artifact_id][0]
```
efredriksson-modelon, line 519: Need to handle the case where [0] fail because you get an empty list.

efredriksson-modelon (changes requested):
Nice I like the design! :watermelon: 

Just a few minor things to fix.

modelon/impact/client/entities/case.py
```diff
@@ -127,6 +129,66 @@ def simulation_log_level(self, simulation_log_level: str):
...
+    """CustomArtifact class"""
+
+    def __init__(
+        self,
+        workspace_id: str,
+        experiment_id: str,
+        case_id: str,
+        data: Dict[str, str],
+        exp_sal,
+    ):
```
iakovn, line 142: Is there any reason to have `Dict` as argument instead of explicit:
```
 def __init__(
        self,
        workspace_id: str,
        experiment_id: str,
        case_id: str,
        exp_sal,
        artifact_id: str,
        download_as: Optional[str] = None
    ):
```

efredriksson-modelon (approved):
:+1:

modelon/impact/client/entities/case.py
```diff
@@ -127,6 +129,66 @@ def simulation_log_level(self, simulation_log_level: str):
...
+    def id(self):
+        """Id of the custom artifact."""
+        return self._data['id']
+
+    @property
+    def download_as(self):
+        """File name for the downloaded artifact."""
+        return self._data['downloadAs']
+
+    def download(self, path: Optional[str] = None):
```
iakovn, line 159: I'd like to see, e.g., `get_artifact` or `get_artifact_data` returning stream as well without writing to file. This is to cover the case of small artifacts like json files that client can easily process in memory. For a multisim ~10000 explicitly downloading all artifacts will be big additional overhead.

modelon/impact/client/entities/case.py
```diff
@@ -430,23 +492,49 @@ def get_trajectories(self) -> Result:
             self._sal,
         )
 
-    def get_artifact(self, artifact_id: str) -> Tuple[bytes, str]:
+    def get_artifact(self, artifact_id: str) -> CustomArtifact:
         """
```
iakovn, line 496: If the user knows artifactId then the extra call to get list of artefacts is unnecessary overhead. For incorrect artifact there'll exception when trying to download. To satisfy both:
```
def get_artifact(self, artifact_id: str, check_id = True)
```
If `check_id` is True the function checks if the artifact is present on the server. If the argument is False no remote calls are done and exception may occur when downloading the artifact.

iakovn (changes requested):
I think we're converging :-) 🚜 

See comments. I'm after supporting mutlisim where each simulation may produce artifacts with the same id/filename. Creating directory to download each case separately seems like large overhead. Generated artifact may be just a small json file.

modelon/impact/client/entities/case.py
abhilash-kumar-nair: @iakovn  I was thinking that if in the future we a add more keys to the response to the GET /custom-artifacts endpoint, it probably might make more sense to keep it as a dict here.

modelon/impact/client/entities/case.py
abhilash-kumar-nair: Ah ok. Would it be ok to have a method to_stream to do this?
so the workflow would look like
```
artifact = case.get_artifact(some_id)
stream=artifact.to_stream()
```

modelon/impact/client/entities/case.py
iakovn: Not really. `download_as` can be None up until we actually need to download. Other alternative is to have:
```
def get_artifact(self, artifact_id: str, download_as = None)
```
and skip calling remote if `download_as` is specified. That would also cover the case of multisim when the client sets different file names when downloading artifacts for each case instead of creating separate directories per case.

modelon/impact/client/entities/case.py
iakovn: I'd prefer `get_data` to indicate that it's not conversion but a request to server and it's not an asyncio stream but a full data object.

modelon/impact/client/entities/case.py
abhilash-kumar-nair: Thanks for the suggestions @iakovn ! I have now pushed the suggested fixes :)

modelon/impact/client/entities/case.py
```diff
@@ -455,16 +536,63 @@ def get_artifact(self, artifact_id: str) -> Tuple[bytes, str]:
...
+            custom_artifact = case.get_artifact(artifact_id)
         """
         assert_successful_operation(self.is_successful(), self._case_id)
-        result, file_name = self._sal.experiment.case_artifact_get(
-            self._workspace_id, self._exp_id, self._case_id, artifact_id
+        if not download_as:
+            resp = self._sal.experiment.case_artifacts_meta_get(
+                self._workspace_id, self._exp_id, self._case_id
+            )
+            meta: List[Dict[str, str]] = [
```
iakovn, line 546: Construct:  `x = [x for x in ... ]` implies different data type for `x` (collection vs single item) which is a bit confusing. I suggest in this case:

`meta = next((meta for meta in resp["data"]["items"] if meta["id"] == artifact_id), None)`

modelon/impact/client/entities/case.py
```diff
@@ -127,6 +129,89 @@ def simulation_log_level(self, simulation_log_level: str):
...
+
+            artifact --
+                The artifact byte stream.
+
+        Example::
+            artifact = case.get_artifact("ABCD")
+            stream = artifact.to_stream()
+            with open(artifact.download_as, "wb") as f:
+                f.write(result)
+        """
```
iakovn, line 207:
```
data = artifact.get_data() # may raise exception on communication error
with open(artifact.download_as, "wb") as f:
    f.write(data)
```

modelon/impact/client/entities/case.py
```diff
@@ -127,6 +129,89 @@ def simulation_log_level(self, simulation_log_level: str):
...
+        if path is None:
+            path = os.path.join(tempfile.gettempdir(), "impact-downloads")
+        os.makedirs(path, exist_ok=True)
+        artifact_path = os.path.join(path, self.download_as)
+        with open(artifact_path, "wb") as f:
+            f.write(artifact)
+        return artifact_path
+
+    def to_stream(self) -> bytes:
+        """
```
iakovn, line 194: `get_data` per previous comment. Streams are really about asyncio and this is not the case.

iakovn (changes requested):
few more small things

iakovn (approved):
thanks for fixing! 🚀

PR #178, opened 2023-03-14
feat: added possibility to filter experiments for models

abhilash-kumar-nair:


tests/impact/client/sal/test_workspace_service.py
```diff
@@ -161,6 +161,12 @@ def test_get_experiments(self, get_all_experiments):
                 ]
             }
         }
+        assert get_all_experiments.adapter.called
+        get_exps_call = get_all_experiments.adapter.request_history[0]
+        assert (
+            f'http://mock-impact.com/api/workspaces/workspace_1/experiments?classPath='
```
efredriksson-modelon, line 167: This does not look right, `?classPath=` should not be in the url when not asking for any specific classpath.

tests/impact/client/sal/test_workspace_service.py
```diff
@@ -161,6 +161,12 @@ def test_get_experiments(self, get_all_experiments):
                 ]
             }
         }
+        assert get_all_experiments.adapter.called
```
efredriksson-modelon, line 164: Can you also add a test where `experiments_get` is called with a class_path?

efredriksson-modelon (changes requested):

abhilash-kumar-nair:
Thanks for the review @efredriksson-modelon . I have made the suggested changes!

abhilash-kumar-nair:
> From the discussion I still think it is unclear what the use case is. Depending on this you could either do it like this or have it on the model class:
> 
> ```
> model = workspace.get_model(...)
> model.get_experiments()
> ```
> 
> If we could have some end-user script on how this feature would be used I think it would become more apparent which solution is most intuitive.

I agree. Thanks for the ideas Emil. I have now made this change. I am not sure, how much will this be helpful for the user.

iakovn:
> From the discussion I still think it is unclear what the use case is. Depending on this you could either do it like this or have it on the model class:
> ```
> model = workspace.get_model(...)
> model.get_experiments()
> ```
> 
> If we could have some end-user script on how this feature would be used I think it would become more apparent which solution is most intuitive.

Let me try to explain the use case a bit more. We currently only have "full forward flow":
```
model = workspace.get_model(...)
experiment_definition = model.new_experiment_definition(dynamic)
do_postprocessing(exp)
```
The missing part is "restarted" flow:
```
model = workspace.get_model(...)
experiment_definition = model.new_experiment_definition(dynamic)
exps = workspace.get_experiments(class_name)
exp = None
for e  in exps:
   if is_this_matching_my_analysis(e, experiment_definition):
      exp = e
      break
if not exp:
do_postprocessing(exp)
```

abhilash-kumar-nair:
> get_experiments

Thanks for the comments @iakovn . I have now moved the method to the workspace entity.

abhilash-kumar-nair:
> From the discussion I still think it is unclear what the use case is. Depending on this you could either do it like this or have it on the model class:
> 
> ```
> model = workspace.get_model(...)
> model.get_experiments()
> ```
> 
> If we could have some end-user script on how this feature would be used I think it would become more apparent which solution is most intuitive.

Hej @efredriksson-modelon , I was not able to implement the above suggestion due to a circular import(See https://jenkins.modelon.com/internal/blue/organizations/jenkins/WAMSTeam%2Fimpact-client-python/detail/experiment-for-model/8/pipeline/ for the failure), so I have now moved the functionality now to the workspace entity.

iakovn (approved):
I'm find with this implementation. I do not have a strong opinion if we should have `model.get_experiments()`, `workspace.get_experiments(class_path)` or both as it seems that different flows maybe relevant depending on the specific workflow.

PR #179, opened 2023-03-14
Build improvements

abhilash-kumar-nair:
General build improvements


Dockerfile
```diff
@@ -1,33 +1,37 @@
-FROM python:3.7
+FROM python:3.9.5
```
iakovn, line 1: I think we agreed on 3.9.6 for a different project and should probably use that here as well.

iakovn (approved):
Please consider switching to Python 3.9.6 based image
Fine otherwise.

PR #184, opened 2023-03-28
fix: corrected upload_fmu uri

abhilash-kumar-nair:
This PR corrected the route for FMU imports.


modelon/impact/client/sal/project.py
```diff
@@ -87,7 +87,7 @@ def fmu_upload(
         url = (
             self._base_uri
             / f"api/workspaces/{workspace_id}/projects/{project_id}/content/"
```
efredriksson-modelon, line 89: And it ´does not contain workspaces/{workspace_id} anymore? That is the bigger change.

efredriksson-modelon (changes requested):
I think there are two issues with the URI :smile:

abhilash-kumar-nair:
Hej @efredriksson-modelon , @iakovn , Could you please take a look at the fmu import changes? Also corrected the ContentType enum which was outdated.

iakovn (approved):
Looking good! 👍

modelon/impact/client/entities/model.py
```diff
@@ -276,3 +281,112 @@ def new_experiment_definition(
...
+        return FMUImportOperation[Model](
+            resp['data']['location'],
+            self._workspace_id,
+            self._project_id,
+            self._sal,
+            Model.from_operation,
+        )
+
+    @classmethod
+    def from_operation(cls, operation: BaseOperation[Model], **kwargs: Any) -> Model:
```
efredriksson-modelon, line 390: Should this method be private?

efredriksson-modelon (approved):
Nice work :+1:

modelon/impact/client/entities/model.py
abhilash-kumar-nair: Yes I think it does. I will push another PR doing this for all the from_operation class methods. Thanks! :)

PR #187, opened 2023-03-30
fix: partial resolution of circular imports

abhilash-kumar-nair:
This PR tries to partly resolve the circular import issues we have in the code base.

### DIFF
--- a/modelon/impact/client/client.py
+++ b/modelon/impact/client/client.py
@@ -10,7 +10,7 @@ import modelon.impact.client.jupyterhub
 from modelon.impact.client.credential_manager import CredentialManager
 from modelon.impact.client.operations.project_import import ProjectImportOperation
 from modelon.impact.client.operations.workspace.imports import WorkspaceImportOperation
-from modelon.impact.client.entities.project import Project, ProjectDefinition, VcsUri
+from modelon.impact.client.entities.project import Project, VcsUri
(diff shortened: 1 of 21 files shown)


iakovn (commented):
As mentioned in the chat I think we should really try to address the core issue rather than try to work that around. I think your change is in the right direction but I'd prefer some more sugar around it per below.

The problem seems to be that entities needs to create operations and operations need to create entities. Additional problem seems to be that `BaseOperation` class `wait` and `data` methods return `Any` while they should really return specific entities. One solution could be to have Operation base class defined as generic and require some callback via protocol as in https://peps.python.org/pep-0544/#callback-protocols , i.e., introduce:
```
Entity = TypeVar("Entity")
class EntityFromOperation(Protocol):
   def __call__(self, operation: BaseOperation[Entity])->Entity:
      ...

class BaseOperation:
   def __init__(self, create_entity: EntityFromOperation):
       self._create_entity = create_entity
   def wait(self) -> Entity:
       pass 
   ...
```
Entity classes would need to implement `create_entity` protocol as, e.g., "from_operation" methods which seems nicer than passing constructors.

The operation classes would then call `self._create_entity(self)` instead of explicitly calling constructors.

The loop will be broken since Operations will not need specific Entity classes. 

Hope this makes some sense. Let's talk further next week.

iakovn (approved):
Good to skip circular imports! 👍

PR #189, opened 2023-04-03
fix: refactored login workflow to distinguish if python client is run…

abhilash-kumar-nair:
As discussed in chat, I have now simplified the workflows to distinguish user using the python client inside vs outside the JH environment. Also made sure server key is always present -  For with JH we rely on env variable JUPYTERHUB_SERVICE_PREFIX else we rely on the response.

### DIFF
--- a/modelon/impact/client/client.py
+++ b/modelon/impact/client/client.py
@@ -149,9 +149,11 @@ class Client:
         self._uri = URI(url)
         self._sal = modelon.impact.client.sal.service.Service(self._uri, context)
 
-        try:
-            self._validate_compatible_api_version()
-        except modelon.impact.client.sal.exceptions.AccessingJupyterHubError:
+        if self._sal.is_jupyterhub_url():
+            logger.info(
+                "API response indicates that the URL '%s' hosts a JupyterHub.",
+                str(self._uri),
+            )
             self._uri, jupyter_context = modelon.impact.client.jupyterhub.authorize(
                 self._uri,
                 interactive,
@@ -161,7 +163,8 @@ class Client:
             self._sal = modelon.impact.client.sal.service.Service(
                 self._uri, jupyter_context
             )
-            self._validate_compatible_api_version()
+
+        self._validate_compatible_api_version()
 
         if credential_manager is None:
             help_hint = f"can be generated at {self._uri / 'admin/keys'}"
--- a/modelon/impact/client/jupyterhub/authorize.py
+++ b/modelon/impact/client/jupyterhub/authorize.py
@@ -1,4 +1,5 @@
 import logging
+import os
 from typing import Optional, Tuple
 from modelon.impact.client.credential_manager import CredentialManager
 from modelon.impact.client.jupyterhub import exceptions
@@ -17,6 +18,10 @@ def _get_jupyter_token(credential_manager: CredentialManager, interactive: bool)
     return jupyter_token
 
 
+def _running_in_jupyterhub_environment() -> bool:
+    return os.environ.get("JUPYTERHUB_SERVICE_PREFIX") is not None
+
+
 def authorize(
     uri: URI,
     interactive: bool,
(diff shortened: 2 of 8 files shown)


modelon/impact/client/client.py
```diff
@@ -149,19 +150,25 @@ def __init__(
...
-            self._validate_compatible_api_version()
-        except modelon.impact.client.sal.exceptions.AccessingJupyterHubError:
-            self._uri, jupyter_context = modelon.impact.client.jupyterhub.authorize(
-                self._uri,
-                interactive,
-                context,
-                jupyterhub_credential_manager,
-            )
+        if self._sal.is_jupyterhub_url():
+            if os.environ.get("JUPYTERHUB_SERVICE_PREFIX") is not None:
```
efredriksson-modelon, line 154: This line is to check if we are inside the notebook or not I assume? It would be nice if this was clarified with a function explain this. So it reads:
```
if running in the jupyter notebook:
   stuff
```

modelon/impact/client/jupyterhub/authorize.py
```diff
@@ -62,3 +63,26 @@ def authorize(
...
+    jupyter_context.token = os.environ.get('JUPYTERHUB_API_TOKEN', '')
+    server = os.environ.get('JUPYTERHUB_SERVICE_PREFIX')
+    user = service.get_user_data(uri, jupyter_context, server)
+
+    if not user.server_running():
+        uri_to_start_server = uri / 'hub/home'
+        raise exceptions.NoJupyterHubServerRunningError(
+            "The user does not have a server running. "
+            f"Go to {uri_to_start_server} and start a server."
+        )
```
efredriksson-modelon, line 86: This is duplicated from the function above.

abhilash-kumar-nair:
Thanks @iakovn . @efredriksson-modelon , I have made the suggested fixes. Could you please take another look at it? Thanks!

modelon/impact/client/client.py
```diff
@@ -149,9 +149,11 @@ def __init__(
         self._uri = URI(url)
         self._sal = modelon.impact.client.sal.service.Service(self._uri, context)
 
-        try:
-            self._validate_compatible_api_version()
-        except modelon.impact.client.sal.exceptions.AccessingJupyterHubError:
+        if self._sal.is_jupyterhub_url():
```
efredriksson-modelon, line 152: One issue with doing this call before the `_validate_compatible_api_version` call is that you do not get the nicer error handling that re-throws the `except modelon.impact.client.sal.exceptions.CommunicationError`. There should probably be some error handling in `is_jupyterhub_url`, so that it always return a boolean and not exceptions.

efredriksson-modelon (changes requested):
The code is easier to read now :+1:

A few potential bugs that should be worked through.

modelon/impact/client/jupyterhub/sal.py
```diff
@@ -73,7 +75,7 @@ def get_user_data(cls, uri: URI, context: JupyterContext) -> JupyterUser:
 
         try:
             user_data = user_response.json()
-            return JupyterUser(user_data['name'], user_data['server'])
+            return JupyterUser(user_data['name'], user_data['server'] or server)
```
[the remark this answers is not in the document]
iakovn: `user_data.get('server', server)`

modelon/impact/client/sal/service.py
```diff
@@ -82,14 +82,18 @@ def wrapped(*args: Any, **kwargs: Any) -> Callable:
             self.external_result, retry_with_login_decorator
         )
 
+    def is_jupyterhub_url(self) -> bool:
+        url = (self._base_uri / "api/").resolve()
```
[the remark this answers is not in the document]
iakovn: I think this is preferable. Otherwise we don't really know if `/api` respose comes from impact or jupyterhub (so far versions are different but that is not reliable). So it should be either 200 -> jupyterhub/true or 404 -> not jupyterhub/false -> othewize comm exception is fine IMO as the answer would be unexpected.

abhilash-kumar-nair:
Thanks @efredriksson-modelon , @iakovn , Fixed it! :)

modelon/impact/client/jupyterhub/authorize.py
```diff
@@ -17,13 +18,27 @@ def _get_jupyter_token(credential_manager: CredentialManager, interactive: bool)
...
     context: Optional[Context] = None,
     credential_manager: Optional[CredentialManager] = None,
     service: Optional[sal.JupyterHubService] = None,
 ) -> Tuple[URI, sal.JupyterContext]:
+
+    jupyter_context = sal.JupyterContext(base=context)
+    service = service or sal.JupyterHubService()
+
+    if _running_in_jupyterhub_environment():
+        jupyter_context.token = os.environ.get('JUPYTERHUB_API_TOKEN', '')
```
iakovn, line 37: The problem that I see here is handling of connection from "other" jupyterhub (like on-prem -> impact.modelon.com). It seems the code will always pick-up "current" Jupyterhub environment and there's no obvious way to override that.

abhilash-kumar-nair:
@efredriksson-modelon , @iakovn I will go ahead and merge this interim PR. The login will be improved in another PR based on the conclusion.

PR #190, opened 2023-04-06
fix: catching general exception while checking for jupyter hub url

abhilash-kumar-nair:


efredriksson-modelon (changes requested):
I think this looks solid :+1: 

Can you also add some tests for `is_jupyterhub_ur`l in tests/impact/client/sal/test_services.py as the function is starting to have a few cases we want to ensure are working?

efredriksson-modelon (approved):
:+1:

PR #191, opened 2023-04-09
Boiler plate code

abhilash-kumar-nair:


iakovn (changes requested):
Good with less code lines! 👍 

I do agree with adding optional constructor arguments for `initialize_from` but I don't think making setters instead of functions returning new instances is the right way. The current design is built around the idea of immutable definitions. All methods are returning new instances and there are not "setter" methods. 

Better names for `initialize_from` would be `with_initialize_from` and you could add methods to retrieve options that are provided in the constructor.

abhilash-kumar-nair:
Thanks @iakovn  for the review. I have made the suggested changes. I will squash the commits before merge to have just one breaking change(intialize_from --> with_initialize_from)

PR #192, opened 2023-04-09
fix: removed clone() method from Worksace entity

abhilash-kumar-nair:
BREAKING CHANGE: clone method in Workpace entity has been removed


anna-lindholm (approved):
Looks good! It is important that we document this properly as we discussed before, but that is outside the scope of this PR.

PR #195, opened 2023-04-18
Api coverage

abhilash-kumar-nair:
This PR contains some fixes for the minor missing API's missed out from the MI REST API

### DIFF
--- a/modelon/impact/client/entities/project.py
+++ b/modelon/impact/client/entities/project.py
@@ -281,6 +281,20 @@ class Project:
             for content in self._project_definition.content
         ]
 
+    def get_content(self, content_id: str) -> ProjectContent:
+        """Gets the project content with the given ID.
+
+        Args:
+            content_id: The ID of the workspace.
+
+        Example::
+
+            project.get_content("79sd8-3n2a4-e3t24")
+
+        """
+        resp = self._sal.project.project_content_get(self.id, content_id)
+        return ProjectContent(resp, self.id, self._sal)
+
     def get_content_by_name(
         self, name: str, content_type: Optional[ContentType] = None
     ) -> Optional[ProjectContent]:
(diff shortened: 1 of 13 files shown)


modelon/impact/client/entities/workspace.py
```diff
@@ -171,6 +173,15 @@ def __repr__(self) -> str:
     def __eq__(self, obj: object) -> bool:
         return isinstance(obj, Workspace) and obj._workspace_id == self._workspace_id
 
+    @property
+    def size(self) -> float:
+        """Workspace size in bytes."""
```
iakovn, line 178: This does not feel like right interface since size of workspace may be changing during the simulation. 
Options are:
1. Always fetch the size in this property and skip constructor argument
2. Use a method `def get_size(refresh=true)` to provide caching support. In that case response from `workspace_get` call should be cached into self._size. `refresh=true` means the size is recomputed.

modelon/impact/client/entities/project.py
```diff
@@ -237,6 +239,20 @@ def id(self) -> str:
...
         return self._project_id
 
+    @property
+    def name(self) -> str:
+        """Project name."""
+        return self.definition.name
+
+    @property
+    def size(self) -> float:
+        """Project size in bytes."""
```
iakovn, line 249: same comment as for workspace size

iakovn (changes requested):
Nice catch that we should expose size information. 👍  I believe we should be more explicit about caching though

modelon/impact/client/entities/workspace.py
abhilash-kumar-nair: Hej @iakovn ,  Good point! I have chosen option 1. :)

iakovn (approved):
looking good! 👍

PR #197, opened 2023-04-27
Interfaces

abhilash-kumar-nair:


iakovn (commented):
Let me summarize my understanding of the problem we're trying to solve and the approach (maybe copy into description?):

We have, for instance, classes like `Model` and `SimpleModelicaExperimentDefinition`.
- `Model` has `new_experiment_definition` method that needs to call `SimpleModelicaExperimentDefinition` constructor.
- `SimpleModelicaExperimentDefinition` has `assert_valid_args` that needs to call `is_instance` on the constructor argument `model`.
This leads to circular import.

Our solution is to introduce `BaseModel` class that is used in the assertion function. The `BaseModel` does not have all the methods but is only used as the base class for the `Model`. As the only implementation of `BaseModel` is the `Model` class assertion text will still report that required argument must be `Model`.

Secondary problem (which really has similar origin) is due to static type checking. For that to work `Model` class is imported conditionally under `if TYPE_CHECKING:` in the experiment module.

@abhilash-kumar-nair  & @efredriksson-modelon do you share this understanding and agree on the pattern?

abhilash-kumar-nair:
> Secondary problem (which really has similar origin) is due to static type checking. For that to work `Model` class is imported conditionally under `if TYPE_CHECKING:` in the experiment module.

Hi @iakovn , Thanks for summarizing the solution nicely. Yes I agree with it :)

efredriksson-modelon:
@abhilash-kumar-nair @iakovn I think adding inheritance with base classes here adds a lot of complexities. You should always be careful with adding inheritance and it seems like this could be more precisely solved with Protocol?

I did a quick spike and pushed up the result to the branch `interfaces-with-protocol` to show what I mean.

iakovn (approved):
👍

efredriksson-modelon (approved):
This is probably a pragmatic way to move forward :+1:

Would be interesting to do some analysis of dependencies and see what can be improved. I tested `pydeps` but it does not handle `if TYPE_CHECKING:` or imports in `__init__.py` files very good.

PR #198, opened 2023-04-27
fix: check if the user given URL is as substring of the IMPACT_URL va…

abhilash-kumar-nair:
Jira - https://modelonproducts.atlassian.net/browse/FLOW-45


iakovn (approved):
👍

PR #200, opened 2023-04-28
fix: default to https://impact.modelon.cloud/ if no url is provided by user or no environmental variable is set

abhilash-kumar-nair:
This PR changes the default url to https://impact.modelon.cloud/ if no url is provided by user or no environmental variable is set


iakovn (approved):
👍

PR #205, opened 2023-05-16
feat: added support to fetch only last point for a result trajectory

abhilash-kumar-nair:
Jira - https://modelonproducts.atlassian.net/browse/FLOW-36

How to test 
```
from modelon.impact.client import Client, Range
import pandas as pd

client = Client(
    url="https://jhmi-staging.modelon.com/user/abhilash.kumar@modelon.com/impact",
    interactive=True,
)
ws = client.get_workspace("tpl")
model = ws.get_model('Unnamed.Structural')
cf = ws.get_custom_function('dynamic')
exp_def = model.new_experiment_definition(cf).with_modifiers({'b': Range(1, 5, 5)})
exp = ws.execute(exp_def).wait()
traj = exp.get_last_time_point(["a", "b", "c"])
print(traj.as_lists())
# Outputs
# [[1.0, 1.0, 1.0], [1.0, 2.0, 2.0], [1.0, 3.0, 3.0], [1.0, 4.0, None], [1.0, 5.0, None]]
print(traj.cases)
# Outputs
# ['case_1', 'case_2', 'case_3', 'case_4', 'case_5']
print(traj.variables)
# Outputs
# ['a', 'b', 'c']
df = pd.DataFrame(data=traj.as_lists(), columns=traj.variables, index=traj.cases)
df.index.name = "Cases"
print(df)
# Outputs
#           a    b    c
# Cases
# case_1  1.0  1.0  1.0
# case_2  1.0  2.0  2.0
# case_3  1.0  3.0  3.0
# case_4  1.0  4.0  NaN
# case_5  1.0  5.0  NaN

```

### DIFF
--- a/modelon/impact/client/entities/experiment.py
+++ b/modelon/impact/client/entities/experiment.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 import logging
 import enum
-from typing import Any, List, Dict, Optional, TYPE_CHECKING
+from typing import Any, List, Dict, Optional, Union, TYPE_CHECKING
 
 from modelon.impact.client.entities.interfaces.experiment import ExperimentInterface
 from modelon.impact.client.operations import experiment
@@ -21,6 +21,49 @@ if TYPE_CHECKING:
     from modelon.impact.client.operations.base import BaseOperation
 
 logger = logging.getLogger(__name__)
+
+ScalarValue = Union[float, int, str]
+
+
+@enum.unique
+class _TrajectoryResponseFormat(enum.Enum):
+    V1 = "application/vnd.impact.trajectories.v1+json"
+    v2 = "application/vnd.impact.trajectories.v2+json"
+
+
+class ExperimentResultPoint:
+    """Value class with the single time instance data in a experiment."""
+
+    def __init__(self, trajectories: List[Dict[str, Any]], variables: List[str]):
+        self._trajectories = trajectories
+        self._variables = variables
+
+    @property
+    def variables(self) -> List[str]:
+        """List of variables."""
+        return self._variables
+
+    @property
+    def cases(self) -> List[str]:
+        """List of case ids."""
+        return [case["caseId"] for case in self._trajectories]
+
+    def as_lists(self) -> List[Optional[List[Optional[ScalarValue]]]]:
+        """Return a list of results per case.
+
+        None indicates that the case was not run or failed. None value
+        for a variable indicated that the variable is not present in the
+        specific case output.
+
+        Example::
+
+            result_data = result.as_lists()
+
+        """
+        return [
+            [item["trajectory"][0] if item else None for item in case_data["items"]]
+            for case_data in self._trajectories
+        ]
 
 
 @enum.unique
@@ -343,6 +386,24 @@ class Experiment(ExperimentInterface):
         """
         return [case for case in self.get_cases() if case.meta.label == case_label]
 
+    def _validate_and_fetch_trajectories(
+        self,
+        variables: List[str],
+        only_last_point: bool = False,
+        format: _TrajectoryResponseFormat = _TrajectoryResponseFormat.V1,
+    ) -> Any:
+        if not isinstance(variables, list):
+            raise TypeError(
+                "Please specify the list of result keys for the trajectories of "
+                "intrest!"
+            )
+        _assert_experiment_is_complete(self.run_info.status, "Simulation")
+        assert_variable_in_result(variables, self.get_variables())
+
+        return self._sal.experiment.trajectories_get(
+            self._workspace_id, self._exp_id, variables, only_last_point, format.value
+        )
+
     def get_trajectories(self, variables: List[str]) -> Dict[str, Any]:
         """Returns a dictionary containing the result trajectories for a list
         of result variables for all the cases.
@@ -366,16 +427,8 @@ class Experiment(ExperimentInterface):
             time = result['case_1']['time']
 
         """
-        if not isinstance(variables, list):
-            raise TypeError(
-                "Please specify the list of result keys for the trajectories of "
-                "intrest!"
-            )
-        _assert_experiment_is_complete(self.run_info.status, "Simulation")
-        assert_variable_in_result(variables, self.get_variables())
-
-        response = self._sal.experiment.trajectories_get(
-            self._workspace_id, self._exp_id, variables
+        response = self._validate_and_fetch_trajectories(
+            variables, only_last_point=False, format=_TrajectoryResponseFormat.V1
         )
 
         if response:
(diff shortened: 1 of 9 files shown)


modelon/impact/client/entities/experiment.py
```diff
@@ -327,12 +327,16 @@ def get_cases_with_label(self, case_label: str) -> List[Case]:
         """
         return [case for case in self.get_cases() if case.meta.label == case_label]
 
-    def get_trajectories(self, variables: List[str]) -> Dict[str, Any]:
+    def get_trajectories(
+        self, variables: List[str], last_point_only: bool = False
+    ) -> Dict[str, Any]:
```
iakovn, line 332: I think from user facing API perspective it's preferable to have a separate method and a different datatype. There's no reason to return "trajectory" objects when we only have a single number. Have you considered returning numpy array or dataframe instead?

iakovn (changes requested):
This will be a good feature! I feel we need to think a bit more about user facing API.

abhilash-kumar-nair:
Hej @iakovn , I have made the discussed fixes now. Could you please take another look? :)

modelon/impact/client/entities/experiment.py
```diff
@@ -374,6 +430,37 @@ def get_trajectories(self, variables: List[str]) -> Dict[str, Any]:
             for j in case_nbrs
         }
 
+    def get_last_time_point(self, variables: List[str]) -> ExperimentResultPoint:
```
iakovn, line 433: I checked with Peter: as this should be valid for "steady state" and other CF as well, let's call the method `get_last_point`.

Is it possible to have default for "variables" as `None` to get all variables?

modelon/impact/client/entities/experiment.py
```diff
@@ -374,6 +430,37 @@ def get_trajectories(self, variables: List[str]) -> Dict[str, Any]:
...
+            result = experiment.get_last_time_point(['h', 'time'])
+
+            # Convert to Pandas data frame
+            df = pd.DataFrame(data=result.as_lists(), columns=result.variables,
+                index=result.cases)
+            df.index.name = "Cases"
+
+        """
+        trajectories = self._validate_and_fetch_trajectories(
+            variables, True, _TrajectoryResponseFormat.v2
```
iakovn, line 460: I suggest using keyword arguments, e.g.,  `only_last_point=True`, to enhance readability.

modelon/impact/client/entities/experiment.py
```diff
@@ -22,6 +22,52 @@
...
+    def as_lists(self) -> List[Optional[List[Optional[ScalarValue]]]]:
+        """Return a list of results per case.
+
+        None indicates that the case was not run or failed. None value
+        for a variable indicated that the variable is not present in the
+        specific case output.
+
+        Example::
+
+            trajectories = result.as_lists()
```
iakovn, line 60: Those are not trajectories, so use something more generic, e.g., `values` or `result_data`, i.e.,

```
   result_data = result.as_lists()
```

modelon/impact/client/entities/experiment.py
```diff
@@ -22,6 +22,52 @@
...
+
+        """
+        return [
+            [
+                case["items"][i].get("trajectory")[0] if case["items"][i] else None
+                for i, _ in enumerate(self._variables)
+            ]
+            for case in self._trajectories
+        ]
```
iakovn, line 69: This kind of loops worth optimizing as they grow a lot for multi-sims, so if I read it right:
```
 [
            [ value["trajectory"][0] if value else None for value in case_data ]
            for case_data in self._trajectories["items"]
] 
```

modelon/impact/client/entities/experiment.py
```diff
@@ -374,6 +430,37 @@ def get_trajectories(self, variables: List[str]) -> Dict[str, Any]:
...
+
+        Raises:
+            OperationNotCompleteError if simulation process is in progress.
+            OperationFailureError if simulation process was cancelled.
+            TypeError if the variable is not a list object.
+            ValueError if trajectory variable is not present in the result.
+
+        Example::
+
+            result = experiment.get_last_time_point(['h', 'time'])
```
iakovn, line 451: `get_last_point` per above

iakovn (changes requested):
Getting close! 
Mostly minor comments

modelon/impact/client/entities/experiment.py
abhilash-kumar-nair: @iakovn  Hmm trajectories is a list of dicts that looks like this, so that **_self._trajectories["items"]_** wont work right?

```
[
            {
                "caseId": "case_1",
                "items": [
                    {"fixed": True, "trajectory": [1.0]},
                    {"fixed": False, "trajectory": [1.0]},
                ],
            },
            {
                "caseId": "case_2",
                "items": [
                    {"fixed": True, "trajectory": [1.0]},
                    {"fixed": False, "trajectory": [2.0]},
                ],
            },
            {
                "caseId": "case_3",
                "items": [
                    {"fixed": True, "trajectory": [1.0]},
                    None,
                ],
            },
        ]
```

abhilash-kumar-nair: I can probably change from  enumerate(self._variables) to range(len())? Also compute len only once to and save it in a variable. Something like this

        variable_list_length = len(self._variables)
        return [
            [
                case["items"][i].get("trajectory")[0] if case["items"][i] else None
                for i in range(variable_list_length)
            ]
            for case in self._trajectories
        ]

modelon/impact/client/entities/experiment.py
abhilash-kumar-nair: Yes it is. Done. thanks!

modelon/impact/client/entities/experiment.py
iakovn: I'd like to avoid repeated `case["items"][i]` in the inner loop. How about:

 ```
 [
            [ item["trajectory"][0] if item else None for value in case_data["items"] ]
            for case_data in self._trajectories
] 
```
BTW: what happens if a case was not run or failed without generating result? Does backend still return "items"? Alternative is that we need additional "if" for `case_data["items"]` presence

File: modelon/impact/client/entities/experiment.py [hunk 7]
```diff
@@ -374,6 +431,51 @@ def get_trajectories(self, variables: List[str]) -> Dict[str, Any]:
...
         }
 
+    def get_last_point(
+        self, variables: Optional[List[str]] = None
+    ) -> ExperimentResultPoint:
+        """Returns a ExperimentResultPoint class for a list of result variables
+        for all the cases.
+
+        Args:
+            variables: An optional list of result variables to fecth
```
iakovn, line 441: spelling :  "to fecth" -> "to fetch"

iakovn (changes requested):
getting closer - couple of more things

modelon/impact/client/entities/experiment.py
abhilash-kumar-nair: Thanks for optimizing it :) I cross checked that "items" is present in the response for a failed compilation too :) Could you please take another look? Thanks!

iakovn (approved):
Lets merge! 🚀

PR #206, opened 2023-05-17
Filter should always be sent as a list

abhilash-kumar-nair:
Filter should always be sent as a list. We were allowing str inputs which will not work.

### DIFF
--- a/modelon/impact/client/options.py
+++ b/modelon/impact/client/options.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 from copy import deepcopy
 from collections.abc import Mapping
-from typing import Union, List, Dict, Any, Iterator
+from typing import List, Dict, Any, Iterator
 from abc import ABC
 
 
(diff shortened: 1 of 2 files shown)


modelon/impact/client/options.py
```diff
@@ -65,26 +65,26 @@ class SolverOptions(BaseExecutionOptions):
...
         Example::
 
             sim_opts = custom_function.get_simulation_options().with_result_filter(
                 pattern = ["*.phi"])
 
         """
-        if not isinstance(pattern, str):
-            pattern = str(pattern)
-        return self.with_values(filter=pattern)
+        if not isinstance(pattern, list):
```
iakovn, line 85: Let's do it all the way: `not isinstance(pattern, list) or [p for p in pattern if not isinstance(p, str)]` and also add test

iakovn (changes requested):
Good improvement! 👍 Let's do it all the way

modelon/impact/client/options.py
abhilash-kumar-nair: Thanks @iakovn ! Fixed it :)

iakovn (approved):
👍

PR #207, opened 2023-05-24
fix: added mapping for python client-MI version compatibility

abhilash-kumar-nair:
https://modelonproducts.atlassian.net/browse/FLOW-24

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -1,7 +1,9 @@
 Install Modelon Impact Client
 =============================
 
-You can install Modelon Impact Client with ``pip``, or by installing from source.
+You can install Modelon Impact Client with ``pip``, or by installing from source. To locate a Python client version
+that works with your version of Modelon Impact, refer to the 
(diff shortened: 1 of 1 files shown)


iakovn (changes requested):
This will be really helpful! Only comment is to use inexact version spec to avoid any confusion about patch versions.
Other thing I'm unsure about is if we need to separate desktop from cloud version

maraberg:
I like this very much. It would be appreciated if this was reflected here as well: https://modelon-impact-client.readthedocs.io/en/latest/install.html

iakovn:
> I like this very much. It would be appreciated if this was reflected here as well: https://modelon-impact-client.readthedocs.io/en/latest/install.html

@maraberg  This will be published when merged. Or do you mean adding a pointer to this text from the first section?

abhilash-kumar-nair:
Hi @maraberg , @iakovn , Thanks for suggestion! I have now made the suggested fixes. Could you please take another look?

maraberg:
Looks good!

iakovn (approved):
👍

PR #213, opened 2023-05-30
chore: bump requests package

abhilash-kumar-nair:


iakovn (changes requested):
It should be "chore" commit as released package requirements are not updated. Or do you think we need to require newer requests in installations?

abhilash-kumar-nair:
@iakovn Semantic bot wont create a release for "chore" I guess? Wanted to bump the python client version in diraction so thought would be good to have the request dependency same as the python client so that I could just do a find and replace in diraction.

abhilash-kumar-nair:
Thanks for the clarification on teams. Fixed it :)

iakovn (approved):
👍

PR #215, opened 2023-06-05
fix: docstring fixes to improve rendering

abhilash-kumar-nair:
This should fix the bad rendering seen in https://modelon-impact-client.readthedocs.io/en/dev/modelon.impact.client.entities.html#modelon.impact.client.entities.case.CaseAnalysis


iakovn (approved):
👍

PR #217, opened 2023-06-05
fix: the Status method in Operations class is now a property

abhilash-kumar-nair:
BREAKING CHANGE: the status method in all operations class inheriting BaseOperation class is now a property


iakovn:
Let's align `status()` to be a property 👍

PR #229, opened 2023-06-13
fix: corrected query string for get_projects api

abhilash-kumar-nair:


iakovn (changes requested):
Could you add a test for this?
On a related topic - have you reviewed test coverage recently? We might have missed some other  minor things as well.

abhilash-kumar-nair:
> Could you add a test for this? On a related topic - have you reviewed test coverage recently? We might have missed some other minor things as well.

Unfortunately we couldn't have caught this one with unit test.. Need integration tests to catch such things since the filtering is done by the server. I can add a integration test. Here's the latest test coverage report - https://jenkins.modelon.com/internal/job/WAMSTeam/job/impact-client-python/job/project-type-fix/1/Test_20coverage_20report/. Its at 90 % but could be improved. :)

abhilash-kumar-nair:
> > Could you add a test for this? On a related topic - have you reviewed test coverage recently? We might have missed some other minor things as well.
> 
> Unfortunately we couldn't have caught this one with unit test.. Need integration tests to catch such things since the filtering is done by the server. I can add a integration test. Here's the latest test coverage report - https://jenkins.modelon.com/internal/job/WAMSTeam/job/impact-client-python/job/project-type-fix/1/Test_20coverage_20report/. Its at 90 % but could be improved. :)

Ah sorry you meant adding a test with a query parameter and checking if the query parameter was sent correctly? That can be done!

iakovn (approved):
👍

PR #231, opened 2023-06-14
fix: only call PUT /case/{case_id} when client state differs from server state

abhilash-kumar-nair:
```

c = Client(url="https://jhmi-staging.modelon.com/", interactive=True)
w = c.get_workspace("test")


m = w.get_model("Unnamed.PID_Controller_1")
ed = m.new_experiment_definition(w.get_custom_function('dynamic'))
e = w.create_experiment(ed)
r = e.execute(with_cases=[]).wait()
cc = e.get_cases()
# print(cc[0]._get_info(cached=False), cc[0].info)
cc[0].meta.label = "bla"
print(cc[0].meta.label)
cc = e.get_cases()
# print(cc[0]._get_info(cached=False))
cc[0].meta.label = "bla"
print(cc[0].meta.label)
print("execute 1")
cc[0].execute().wait()
print("execute 1 done")
cc[0].get_result("csv")[1]
print("execute 2")
cc[0].meta.label = "blah2"
cc[0].execute().wait()
print(cc[0].meta.label)
print("execute 2 done")
print(cc[0].meta)
print("execute 3")
cc[0].execute().wait()

```

### DIFF
--- a/modelon/impact/client/entities/case.py
+++ b/modelon/impact/client/entities/case.py
@@ -23,6 +23,29 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
+def _compare_nested_dicts(
+    dict1: Dict[str, Any], dict2: Dict[str, Any], skip_keys: List[str]
+) -> bool:
+    for key in dict1:
+        if key in skip_keys:
+            continue
+
+        if key not in dict2:
+            return False
+
+        val1 = dict1[key]
+        val2 = dict2[key]
+
+        if isinstance(val1, dict) and isinstance(val2, dict):
+            if not _compare_nested_dicts(val1, val2, skip_keys):
+                return False
+        else:
+            if val1 != val2:
+                return False
+
+    return True
+
+
 def _assert_case_is_complete(
     status: CaseStatus, operation_name: str = "Operation"
 ) -> None:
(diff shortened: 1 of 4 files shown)


modelon/impact/client/entities/case.py
```diff
@@ -744,9 +762,21 @@ def sync(self) -> None:
...
+        skip_keys = [
+            "fmu_id",
+            "analysis_function",
+            "structural_parametrization",
+            "fmu_base_parametrization",
+            "run_info",
+            "id",
+        ]
+        info["run_info"] = server_info.get("run_info", {})
+        if not _compare_nested_dicts(info, server_info, skip_keys):
```
efredriksson-modelon, line 776: I think it would be simpler to skip this and always call case_put. The performance is not critical here and it adds a lot of code. In the future we will probably have a PATCH method also would change the code in the future so adding a more complicated solution is probably not worth it.

efredriksson-modelon (changes requested):
Good initiative! :green_apple:  (and sorry for the slow review)

The idea is for fixing this for now is sound but it is probably best to go for the simpler solution.

modelon/impact/client/entities/case.py
iakovn: I feel like we should avoid all that complexity and just add the PATCH end-point into API.

Other aspect I'm not sure about is if it's really worth it to always do `sync()` (i.e., PUT/PATCH meta) before each execute unconditionally. We could instead introduce some flag `_is_updated` that is set in each property setter method and only PUT/PATCH if that flag is set. Same flag may need checking in "get_info" to avoid override. This is relevant if:
- User sets, e.g., a label in a script and calls `get_info` (no cache) that overrides the label
- A user is switching between UI and notebook, e.g., setting label via UI which can get override from script

PR #233, opened 2023-06-15
chore: bump supported MI version

abhilash-kumar-nair:


iakovn (approved):
We should find some more generic way... But good for now

PR #241, opened 2023-12-05
Add method to rename workspace

abhilash-kumar-nair:


modelon/impact/client/entities/workspace.py
```diff
@@ -407,6 +413,20 @@ def size(self) -> float:
     def definition(self) -> WorkspaceDefinition:
         return self._workspace_definition
 
+    @property
+    def name(self) -> str:
```
iakovn, line 417: This property becomes a bit weird since setting the name results in a remote call and definition re-fetch while reading returns client-side value and does not make additional server request. This implies:
```
workspace = client.get_workspace(id)
same_workspace = client.get_workspace(id)
workspace.name = "new_name"
assert workspace.name != same_workspace.name
same_workspace = client.get_workspace(id)
assert workspace.name == same_workspace.name
```
Therefore, I this I'd prefer a "rename" method with docstring clearly stating that it does a remote call.

modelon/impact/client/entities/workspace.py
```diff
@@ -407,6 +413,20 @@ def size(self) -> float:
     def definition(self) -> WorkspaceDefinition:
```
iakovn, line 413: Please add a docstring and indicate that this is a client side object (may be off-sync with the server).

iakovn (changes requested):
I'm not very comfortable with "name" as a property. Let's discuss or change if you agree. There should also be docstring on a public method.

modelon/impact/client/entities/workspace.py
abhilash-kumar-nair: I agree. I have now added a explicit rename method. However I think we should hide details like "the api does a remote" from the user. Instead I think we should always try to fetch the definition when workspace.definition is called. If we do this we can avoid such weird behavior like the one you pointed out. and always be confident of being in sync with the remote? I think the cost of doing the call to remote always isn't significant. Thoughts? :)

modelon/impact/client/entities/workspace.py
abhilash-kumar-nair: Hmm I think we should aim to never let the user worry about not being in sync.. but I am open to discussion :)

iakovn (approved):
Let's do this! 👍

PR #242, opened 2023-12-13
fix: corrected the api path for request access

abhilash-kumar-nair:


iakovn:
LGTM :-) Don't know if it's worth fixing just that or you could add "grant" and some function to list accesses as well in the same PR

iakovn (approved):
LGTM :-) Don't know if it's worth fixing just that or you could add "grant" and some function to list accesses as well in the same PR

abhilash-kumar-nair:
Thanks @iakovn  for the review! Sure. I can add that in a separate MR once the workspace service PR is in. I wanted this one in quick to not have this method broken once we deploy the new workspace service :)

PR #243, opened 2023-12-15
New published client

abhilash-kumar-nair:


iakovn (changes requested):
It'll be a great feature! 
I think we should keep "@Experimental" on all the peer-to-peer methods for the coming release but remove experimental from the "find" method.
We should also double-check that `get_by_id` returns 404 for people without access and does not allow to determine if the specified ID exists or not in any way. Seems like it's a rather general recommendation, e.g., here https://stackoverflow.com/questions/28582830/access-denied-403-or-404 . Kind of same, e.g., on github/gitlab navigating to a "private" repo URL you get "404" and can't tell if URL is wrong or you just have no access.

abhilash-kumar-nair:
> eer-to-peer methods for the coming release but remove experimental from the "find" method.

I agree. Lets do that once we can are ready :)
I will merge this one and test the get_by_id method.

PR #251, opened 2024-02-01
feat: added support to grant community access to published workspacce

abhilash-kumar-nair:
Jira https://modelonproducts.atlassian.net/browse/FLOW-207


iakovn (approved):
Let's put this in! 👍 
This is experimental for now and we'll discuss further if ACL structures should change

PR #255, opened 2024-02-14
feat: added possibility to export a workspace without group share

abhilash-kumar-nair:
Code to test

```
import os

os.environ["IMPACT_PYTHON_CLIENT_EXPERIMENTAL"] = "true"
from modelon.impact.client import AccessSettings, Client

client = Client(url="https://jhmi-staging.modelon.com", interactive=True)

pwc = client.get_published_workspaces_client()
workspace = client.get_workspace('abhilash')

# Export without group share
workspace.export(publish=True, access=AccessSettings(group_names=[]))

# Export with default group share
workspace.export(
    publish=True, access=AccessSettings(group_names=['impact-tenant-mycompany'])
)

```

### DIFF
--- a/modelon/impact/client/sal/workspace.py
+++ b/modelon/impact/client/sal/workspace.py
@@ -66,10 +66,16 @@ class WorkspaceService:
         return self._http_client.get_json(url)
 
     def workspace_export_setup(
-        self, workspace_id: str, publish: bool, class_path: Optional[str] = None
+        self,
+        workspace_id: str,
+        publish: bool,
+        class_path: Optional[str] = None,
+        access_settings: Optional[Dict[str, Any]] = None,
     ) -> Dict[str, Any]:
         url = (self._base_uri / "api/workspace-exports").resolve()
-        body = {"workspaceId": workspace_id, 'publish': publish}
+        body = {"workspaceId": workspace_id, "publish": publish}
+        if access_settings:
+            body["access"] = access_settings
         if class_path:
             body['appMode'] = {'model': class_path}
         return self._http_client.post_json(url, body=body)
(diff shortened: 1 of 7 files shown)


modelon/impact/client/sal/workspace.py
```diff
@@ -66,10 +66,18 @@ def dependencies_get(
...
+        publish: bool,
+        class_path: Optional[str] = None,
+        access_settings: Optional[Dict[str, Any]] = None,
     ) -> Dict[str, Any]:
         url = (self._base_uri / "api/workspace-exports").resolve()
-        body = {"workspaceId": workspace_id, 'publish': publish}
+        body = {
+            "workspaceId": workspace_id,
+            "publish": publish,
+            "access": access_settings,
```
iakovn, line 79: I suggest we follow the pattern below for class-path and only add optional fields if they're not None. 
So:
```
if access_settings:
  body["access"] = access_settings
```

iakovn (changes requested):
Looking good! 👍 
One minor change only and OK to merge after that!

modelon/impact/client/sal/workspace.py
abhilash-kumar-nair: The access_setting parameter would never be None(See https://github.com/modelon-community/impact-client-python/pull/255/files#diff-138c49731943a8bdb4b84b87ba6de4e5bc54993f304e31903e7462b3604a67cdR713). I felt that adding additional checks to see if groups names key in the access_setting would be not be worth to add in the client as we might add other keys as well in the future?

iakovn: Ok, you're right. I guess I find code like below a bit difficult to read:
```
  access_settings = {
            "groupNames": access.group_names
            if access and isinstance(access.group_names, list)
            else None
        }
```
I'd prefer
```
   if access:
        access_settings = {
            "groupNames": access.group_names # No need for "if hasattr(access, "group_names") else None"
        }
   else:
      access_settings = None
```
but I'm ok with your current implementation.

iakovn (approved):

modelon/impact/client/sal/workspace.py
abhilash-kumar-nair: I agree, this makes it clearer. I have made this change and pushed now! :)

PR #257, opened 2024-02-19
Group grant access

abhilash-kumar-nair:
This PR adds support for granting and revoking group access to a published workspace


iakovn (commented):
Looking good! Could you add some description to the PR before merging?

PR #259, opened 2024-02-29
fix: updated experimental get_access_control_list method as per new protocol on MI

abhilash-kumar-nair:
Code to test
```
import os
os.environ["IMPACT_PYTHON_CLIENT_EXPERIMENTAL"] = "true"
from modelon.impact.client.client import Client

c = Client("https://impact.local")
pwc = c.get_published_workspaces_client()
published_workspace = pwc.get_by_id(sharing_id="e7fd7608-f1b3-48e8-b04e-4bb473f4d57d")

published_workspace.grant_community_access()
published_workspace_acl = published_workspace.get_access_control_list()
print(published_workspace_acl)
```


iakovn (approved):
make more sense now!

PR #260, opened 2024-03-05
Migrating tests to VCRpy

abhilash-kumar-nair:
This PR is a first cut implementation to migrate the mock based tests to vcr based tests.


iakovn (commented):
Great work! This will simplify test setup a lot when it works! 🍉 

Some initial reactions:
- we need some into about VCR setup in the top level README
- I'm not comfortable with `_clean_workspace_and_its_projects(client: Client):` wiping everything in my user account for the recording. Can we instead explicitly work only with specific names of workspaces and projects, e.g., CLIENT_TEST_WORKSPACE_NAME_NNN, and only wipe out those from the account? This would mean that recording should do some cleaning on get_workspaces/get_projects . Not sure how easy it'll be.
- I see actual e-mail to still be present in some yaml files. Can you regexp replace that when doing scrubbing instead of replacing specific fields in json response?

abhilash-kumar-nair:
> * we need some into about VCR setup in the top level README

Thanks @iakovn for the review! I have now made the suggested updates

Makefile
```diff
@@ -2,6 +2,11 @@
 USER_ID := $(shell id -u)
 GROUP_ID := $(shell id -g)
 IN_DOCKER_IMG := $(shell test -f /.dockerenv && echo 1 || echo 0)
```
iakovn, line 4: I was able to run tests on cloud from user space. I suggest adding: 
```
IN_DOCKER_IMG := $(shell ( test -f /.dockerenv && echo 1 ) || ( test -f /opt/impact/notebook_manifest.json && echo 1 ) || echo 0)
```
to enable that out of the box

Makefile
```diff
@@ -2,6 +2,11 @@
...
+SECRETS := $(CURDIR)/.secrets
+export MODELON_IMPACT_CLIENT_API_KEY ?= $(shell cat $(SECRETS)/api.key)
+export JUPYTERHUB_API_TOKEN ?= $(shell cat $(SECRETS)/jupyterhub-api.key)
+export MODELON_IMPACT_CLIENT_URL ?= https://jhmi-staging.modelon.com/
+export MODELON_IMPACT_USERNAME ?= $(shell git config user.email)
```
iakovn, line 9: Please add `export MODELON_IMPACT_CLIENT_INTERACTIVE:=false`  since accidental "true" that I had overrides user's api.key file.

`export MODELON_IMPACT_USERNAME ?= $(shell git config user.email | tr '[:upper:]' '[:lower:]')`
I have apparently capital letters returned from `git config user.email` not matching lowercase in username which prevented cleaning from e-mail from recording. We should add a test that username was correctly set at some point... We could also make replacement not case sensitive.

Makefile
```diff
@@ -53,7 +58,32 @@ experimental-test:
 
 test: build lint
 	IMPACT_PYTHON_CLIENT_EXPERIMENTAL=0 $(MAKE) unit-test
-	IMPACT_PYTHON_CLIENT_EXPERIMENTAL=1 $(MAKE) experimental-test 
+	IMPACT_PYTHON_CLIENT_EXPERIMENTAL=1 $(MAKE) experimental-test
```
iakovn, line 61: I would prefer:
```
unit-test: export IMPACT_PYTHON_CLIENT_EXPERIMENTAL=0
unit-test:
	$(call _run_bare, poetry run -- pytest -vv -m 'not (experimental)' ${EXTRA_PYTEST_FLAGS})

experimental-test: export IMPACT_PYTHON_CLIENT_EXPERIMENTAL=1
experimental-test:
	$(call _run_bare, poetry run -- pytest -vv ${EXTRA_PYTEST_FLAGS})

test: build lint unit-test experimental-test
```
Or are there reasons to run unit-test target in experimental mode?

iakovn (commented):
Few minor things in Makefile and good to go!

I tested running test in vscode client directly from userspace and it worked nicely. The only issue is that responses for `/impact/api/workspaces` contain all workspaces when recording - so I'd need to do some cleaning if I would record things. I don't think this is a blocking issue though,

Makefile
abhilash-kumar-nair: Good catch. I made the replacement case insensitive. :)

abhilash-kumar-nair:
> Few minor things in Makefile and good to go!
> 
> I tested running test in vscode client directly from userspace and it worked nicely. The only issue is that responses for `/impact/api/workspaces` contain all workspaces when recording - so I'd need to do some cleaning if I would record things. I don't think this is a blocking issue though,

Thanks @iakovn for testing and review! I have made the suggested changes now. We might need to add some more "scrub" cases to clean up `/impact/api/workspaces` if required in the future

iakovn (approved):
LGTM! ✈️

PR #262, opened 2024-03-25
Fix fixture

abhilash-kumar-nair:
This PR fixes a minor issue when a user tries regenerate cassettes for just a single test. In this case, although only the single test cassette is regenerated, the assert for username errors out as the user name doesn't match the one in cassettes


iakovn (approved):
Nice catch! 👍

PR #263, opened 2024-03-26
feat: added client method to fetch user data

abhilash-kumar-nair:
This PR adds support for fetching user data using a method on the client class.

### DIFF
--- a/modelon/impact/client/client.py
+++ b/modelon/impact/client/client.py
@@ -725,3 +725,50 @@ class Client:
 
         """
         return PublishedWorkspacesClient(self._sal)
+
+    def get_me(self) -> User:
+        """Return the User class object for the logged in user.
+
+        Example::
+
+            user = client.get_me()
+            user_tenant_group_name = user.tenant.group_name
+
+        """
+        user = self._sal.users.get_me()["data"]
+        return User.from_dict(user)
+
+
+@dataclass
+class Tenant:
+    id: str
+    group_name: str
+
+    @classmethod
+    def from_dict(cls, tenant_data: Dict[str, Any]) -> Tenant:
+        return Tenant(id=tenant_data["id"], group_name=tenant_data["groupName"])
+
+
+@dataclass
+class User:
+    id: str
+    username: str
+    firstname: str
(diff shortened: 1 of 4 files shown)


modelon/impact/client/client.py
```diff
@@ -725,3 +725,50 @@ def get_published_workspaces_client(self) -> PublishedWorkspacesClient:
...
+    id: str
+    group_name: str
+
+    @classmethod
+    def from_dict(cls, tenant_data: Dict[str, Any]) -> Tenant:
+        return Tenant(id=tenant_data['id'], group_name=tenant_data['groupName'])
+
+
+@dataclass
+class User:
```
iakovn, line 753: Any potential benefit if we used a pydantic model instead?

iakovn (commented):
Nice little feature! 🍒 

Some questions on tech used

tests/impact/client/test_client.py
```diff
@@ -294,3 +294,18 @@ def test_get_executions_for_workspace(executions, model_compile, experiment_exec
     opeartions = list(client.get_executions(IDs.WORKSPACE_SECONDARY))
     assert len(opeartions) == 1
     assert isinstance(opeartions[0], ExperimentOperation)
+
+
+def test_get_me(user):
```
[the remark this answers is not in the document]
abhilash-kumar-nair: I did try a VCR implementation for this by setting the "scrubbed" response_body to the dummy dict used in this fixture in this MR but the tests would fail during the regeneration as the responses are only scrubbed before saving to disk, so while generating the cassette it would try to match with actual response body sent by the server. IAlso it seemed like an overkill for this one specific test.

modelon/impact/client/client.py
abhilash-kumar-nair: Probably only for a utility funtion for parsing the dict, we would still need to define the structure(autogenerting from openapi is an option but would contain camelCase attributes). Also we would need to add a pydantic dependency just for this class if we wish to use pydantic.

iakovn (approved):
LGTM 👍

PR #265, opened 2024-04-01
Login flow update[Breaking Change]

abhilash-kumar-nair:
This PR adds support for the new API gateway workflow. This is a breaking change and the commits would be squashed and reworded as a Breaking Change before merging.


abhilash-kumar-nair:
@iakovn , I have now updated this MR with docs and api changes required to migrate to the new login flow. There a few things to consider next:-

- We need to bump the lower bound for the open-api version to ensure, the new version of client fails early against old MI with good error messages.
- Add tags for the old login flow version in RTD
- Update the compatibility matrix

docs/source/authentication_with_modelon_impact.rst
```diff
@@ -24,63 +18,23 @@ in the JupyterHub environment:
     >>>> client = Client()
 
 
-JupyterHub authentication
-*************************
+Authentication outside JupyterHub environment
+*********************************************
```
iakovn, line 23: I suggest merging with the previous section and stop mentioning JupyterHub which is a pure internal component now. We can still be talking "from inside Modelon Impact (either custom functions, scripts or notebooks) vs from outside.

docs/source/authentication_with_modelon_impact.rst
```diff
@@ -4,16 +4,10 @@ Authentication with Modelon Impact
...
-installation, one can create a connection to the external URL. In that case, this will be automatic
-based on the automatically populated ``JUPYTERHUB_API_TOKEN`` environment variable.
-
-An API key must be used in addition to the ``JUPYTERHUB API TOKEN`` to authenticate against the Modelon 
+The cloud/on-premise architecture for Modelon Impact uses an API key to authenticate against the Modelon 
 Impact server operating inside the JupyterHub environment. If the client library is used within the
-JupyterHub environment, this is likewise handled automatically based on the key stored at 
-``~/.impact/api.key`` in that environment.
+JupyterHub environment, this is likewise handled automatically based on the key stored in the environmental
+variable ``MODELON_IMPACT_CLIENT_API_KEY``.
```
iakovn, line 10: Some rewrite and may need to iterate:

In order to work with Modelon Impact REST API a non-browser code needs to pass  `impact-api-key: <api key>` header with all requests. The Python `Client()` constructor will set this header using the value of API key from either (in order of priority):
- explicit constructor parameter or interactively given by a user in interactive mode;
- environment variable `MODELON_IMPACT_CLIENT_API_KEY`;
- `$HOME/.impact/api.key` file.

The `MODELON_IMPACT_CLIENT_API_KEY` environment variable is automatically set with a temporary user specific API key inside the Modelon Impact environment. Therefore explicit user input is not necessary when the Python client is used in a script, notebook or custom function running within the Modelon Impact by a logged in user.

When using Python client from outside Modelon Impact, e.g., from a desktop, the API key needs to be generated via user interface.

modelon/impact/client/client.py
```diff
@@ -240,52 +238,34 @@ def __init__(
...
 
-        self._sal.add_login_retry_with(api_key)
-
-        resp = self._sal.users.get_me()
-        if "license" not in resp["data"]:
-            raise exceptions.NoAssignedLicenseError(
-                "No assigned license during login. Either out of floating Deployment "
-                "Add-on licenses or your assigned user license could not be validated"
-            )
+        # TODO Update to use the unprotected API route https://impact.modelon.cloud/api
```
iakovn, line 267: Any update on this?

iakovn (changes requested):
Some suggestions for documentation updates.

docs/source/authentication_with_modelon_impact.rst
abhilash-kumar-nair: Thanks Iakov for the nice write up. I have now made the suggested changes with some minor tweaks to the order of preference. Please let me know if you find this reasonable.

iakovn:
@abhilash-kumar-nair  please review and merge my docs PR. I'm ok with merging to master and integrating after that. Thanks!

PR #269, opened 2024-04-08
Utility methods for publish

abhilash-kumar-nair:
This PR adds the following utility methods -
1. Fetch corresponding remote published workspace if it exists for a local workspace
published_ws = workspace.get_published_workspace()
2. Check if remote updates are available for a local workspace linked to a published workspace
workspace.has_remote_updates()

### DIFF
--- a/modelon/impact/client/entities/workspace.py
+++ b/modelon/impact/client/entities/workspace.py
@@ -15,6 +15,10 @@ from modelon.impact.client.entities.interfaces.workspace import WorkspaceInterfa
 from modelon.impact.client.entities.model import Model
 from modelon.impact.client.entities.model_executable import ModelExecutable
 from modelon.impact.client.entities.project import Project, VcsUri
+from modelon.impact.client.exceptions import (
+    NoAssociatedPublishedWorkspaceError,
+    RemotePublishedWorkspaceLinkError,
+)
 from modelon.impact.client.experiment_definition.interfaces.definition import (
     BaseExperimentDefinition,
 )
@@ -31,6 +35,7 @@ from modelon.impact.client.operations.workspace.exports import (
     WorkspaceExportOperation,
 )
 from modelon.impact.client.operations.workspace.imports import WorkspaceImportOperation
+from modelon.impact.client.sal.exceptions import HTTPError
 from modelon.impact.client.sal.service import Service
 
 if TYPE_CHECKING:
@@ -445,6 +450,89 @@ class ReceivedFrom:
     @property
     def created_at(self) -> int:
         return self._data["createdAt"]
+
+
+class ReceivedFromWorkspace:
+    def __init__(self, workspace: Workspace) -> None:
+        self._workspace = workspace
+        self._received_from = self._workspace.definition.received_from
+        if not self._received_from:
+            raise NoAssociatedPublishedWorkspaceError(
+                "The workspace has no link to a published workspace."
+            )
+
+    @property
+    def metadata(self) -> ReceivedFrom:
+        """Reference metadata for the published workspace."""
+        assert self._received_from
+        return self._received_from
+
+    def get_workspace(self) -> Optional[PublishedWorkspace]:
+        """Return the published workspace from which the workspace was imported from.
+        Returns None if workspace is not linked to any published workspace.
+
+        Returns:
+            An PublishedWorkspace class object.
+
+        Example::
+
+            published_ws = workspace.received_from.get_workspace()
+
+        """
+        try:
+            sharing_id = self.metadata.sharing_id
+            data = self._workspace._sal.workspace.get_published_workspace(sharing_id)
+        except HTTPError as e:
+            logger.warning(
+                f"Published workspace with ID: {sharing_id} doesn't exist. Cause{e}"
+            )
+            return None
+        definition = PublishedWorkspaceDefinition.from_dict(data)
+        return PublishedWorkspace(
+            data["id"], definition=definition, service=self._workspace._sal
+        )
+
+    def has_updates(self) -> Optional[bool]:
+        """Return True if there are updates available for the published workspace
+        corresponding to the local workspace(if any) else False.
+
+        Example::
+
+            has_updates = workspace.received_from.has_updates()
+
+        """
+        pw = self.get_workspace()
+        if not pw:
+            return None
+        local_ws_created_at = self.metadata.created_at  # type: ignore
+        return pw.created_at != local_ws_created_at
+
+    def _import_to_userspace(self, sharing_id: str) -> Workspace:
+        resp = self._workspace._sal.workspace.import_from_cloud(
+            sharing_id, self._workspace.id
+        )
+        return WorkspaceImportOperation[Workspace](
+            resp["data"]["location"],
+            self._workspace._sal,
+            Workspace.from_import_operation,
+        ).wait()
+
+    def update_userspace(self) -> Workspace:
+        """Returns the workspaces updated from the latest published one.
+
+        Example::
+
+            updated_workspace = workspace.received_from.update_userspace()
+
+        """
+        logger.info("Updating the workspace to the latest published workspace.")
+        pub_ws = self.get_workspace()
+        if not pub_ws:
+            raise NoAssociatedPublishedWorkspaceError(
+                "No published workspace found that are "
+                f"associated with local workspace with ID: {self._workspace.id}"
+            )
+        return self._import_to_userspace(sharing_id=pub_ws.id)
 
 
 class WorkspaceDefinition:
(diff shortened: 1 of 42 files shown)


aldrigblivastur:
Looks good.

But I believe there is one issue: The method `get_published_workspace` assumes that it is being called by the same user that published the workspace. Code that belongs to a published workspace will likely be run by a different user.

Instead of using the method `get_me`, it seems we should be reading out the user that published the workspace via the `ReceivedFrom` property.

But it seems even better to skip the user and just use the unique _sharing id_ that is also found in the `ReceivedFrom` property?
`self._sal.workspace.get_by_id(self.definition.received_from.sharing_id)`

abhilash-kumar-nair:
> self._sal.workspace.get_by_id(self.definition.received_from.sharing_id)

Good point Erik! I agree this makes more sense. I had thought of the _get_published_workspace_ to be used for fetching the PublishedWorkspace class which was published for the corresponding workspace. But the use case you have in mind is more useful. I have updated the implementation, docs and tests accordingly. Could you please take another look?

iakovn:
> > self._sal.workspace.get_by_id(self.definition.received_from.sharing_id)
> 
> Good point Erik! I agree this makes more sense. I had thought of the _get_published_workspace_ to be used for fetching the PublishedWorkspace class which was published for the corresponding workspace. But the use case you have in mind is more useful. I have updated the implementation, docs and tests accordingly. Could you please take another look?

It seems there's some confusion and I think we really need 3 methods (1 for publisher and 2 for receiver):
1. Publisher checks if workspace was published and gets a published workspace object if that exists (so "find" with his own user name and current workspace name, optional model name). Raise excepion if the workspace has "received_from" (workspace should be renamed first which will remove "received_from").
2. Receiver checks if there is an update for the workspace he received getting a boolean flag. Return "false" if published workspace was deleted. 
3. Receiver gets the published workspace his workspace was imported from (return None if deleted by publisher).

For the (2) and (3) it should be some exception if the workspace has no "received_from" info.

Thoughts?

aldrigblivastur:
> It seems there's some confusion and I think we really need 3 methods (1 for publisher and 2 for receiver):
> 
> 1. Publisher checks if workspace was published and gets a published workspace object if that exists (so "find" with his own user name and current workspace name, optional model name). Raise excepion if the workspace has "received_from" (workspace should be renamed first which will remove "received_from").
> 2. Receiver checks if there is an update for the workspace he received getting a boolean flag. Return "false" if published workspace was deleted.
> 3. Receiver gets the published workspace his workspace was imported from (return None if deleted by publisher).
> 
> For the (2) and (3) it should be some exception if the workspace has no "received_from" info.

Ah. Yes. I agree with all of Iakov's comments.

I didn't realize that there is also a use case for a user who publishes the workspace. I focused only on the receiver use cases: checking if there is an update and fetching it. And as a receiver, my only reason to get the published workspace would be to fetch the update.

You could even take that one step further and introduce a fourth method, for fetching an update without having to first get a published workspace object.

```
if workspace.has_remote_updates():
   workspace.update_from_remote()
```

abhilash-kumar-nair:
Thanks @iakovn , @aldrigblivastur  for the discussion! I have now made the suggested changes! :)

modelon/impact/client/entities/workspace.py
```diff
@@ -1126,3 +1210,53 @@ def from_conversion_operation(
...
+        """Returns the ReceivedFromWorkspace class."""
+        return ReceivedFromWorkspace(self)
+
+    def get_published_workspace(
+        self, model_name: Optional[str] = None
+    ) -> Optional[PublishedWorkspace]:
+        """Returns the uploaded published workspace.
+
+        Args:
+            model_name: Class path of the modelica model. If specified,
```
iakovn, line 1225: Class path of the **M**odelica model.

modelon/impact/client/entities/workspace.py
```diff
@@ -1126,3 +1210,53 @@ def from_conversion_operation(
...
+
+    @property
+    def received_from(self) -> ReceivedFromWorkspace:
+        """Returns the ReceivedFromWorkspace class."""
+        return ReceivedFromWorkspace(self)
+
+    def get_published_workspace(
+        self, model_name: Optional[str] = None
+    ) -> Optional[PublishedWorkspace]:
+        """Returns the uploaded published workspace.
```
iakovn, line 1222: Returns the published workspace with the name matching this workspace.

modelon/impact/client/entities/workspace.py
```diff
@@ -1126,3 +1210,53 @@ def from_conversion_operation(
...
+
+            published_ws = workspace.get_published_workspace()
+
+        """
+        recieved_from = self.definition.received_from
+        if recieved_from:
+            raise RemotePublishedWorkspaceLinkError(
+                "The workspace is linked to a published workspace with ID:"
+                f"{recieved_from.sharing_id}. Please rename the workspace to unlink"
+                "it from the published workspace, if you wish to publish it!"
```
iakovn, line 1242: "This workspace is imported from published workspace with ID:"
 f"{recieved_from.sharing_id}. Use received_from.get_workspace() to retrieve the information" 

Do we have workspace rename functionality in client? Anyway, I don't think we need to talk about that here.

modelon/impact/client/entities/workspace.py
```diff
@@ -1126,3 +1210,53 @@ def from_conversion_operation(
...
+            logger.warning(
+                "No published workspace corresponding to the local workspace with"
+                f" ID: {self.id} exist!"
+            )
+            return None
+        return PublishedWorkspace(
+            data[0]["id"],
+            definition=PublishedWorkspaceDefinition.from_dict(data[0]),
+            service=self._sal,
+        )
```
iakovn, line 1262: In the app mode case: can you have several workspaces for different models for the same workspace name? I think you need to check for matching "model"

modelon/impact/client/entities/workspace.py
```diff
@@ -446,6 +451,85 @@ def created_at(self) -> int:
...
+
+            published_ws = workspace.received_from.get_workspace()
+
+        """
+        recieved_from = self._workspace.definition.received_from
+        if not recieved_from:
+            raise NoAssociatedPublishedWorkspaceError(
+                "No published workspace found that are "
+                f"associated with local workspace with ID: {self._workspace.id}"
+            )
```
iakovn, line 475: Seems more logical to have this check before the object gets created ( in the constructor) and not here.

I think a shorter message is sufficient: "The workspace has no link to a published workspace."

modelon/impact/client/entities/workspace.py
```diff
@@ -1126,3 +1210,53 @@ def from_conversion_operation(
     ) -> Workspace:
         assert isinstance(operation, WorkspaceConversionOperation)
         return cls(**kwargs, service=operation._sal)
+
+    @property
+    def received_from(self) -> ReceivedFromWorkspace:
+        """Returns the ReceivedFromWorkspace class."""
+        return ReceivedFromWorkspace(self)
```
iakovn, line 1217: Should be returning None if there is not received_from workspace.

modelon/impact/client/entities/workspace.py
```diff
@@ -1126,3 +1210,53 @@ def from_conversion_operation(
     ) -> Workspace:
         assert isinstance(operation, WorkspaceConversionOperation)
         return cls(**kwargs, service=operation._sal)
+
+    @property
+    def received_from(self) -> ReceivedFromWorkspace:
+        """Returns the ReceivedFromWorkspace class."""
```
iakovn, line 1216: Returns an instance of ReceivedFromWorkspace class.

iakovn (changes requested):
Few minor changes and we're good to go... 🚋

modelon/impact/client/entities/workspace.py
```diff
@@ -446,6 +451,85 @@ def created_at(self) -> int:
...
+
+        Returns:
+            An PublishedWorkspace class object.
+
+        Example::
+
+            published_ws = workspace.received_from.get_workspace()
+
+        """
+        recieved_from = self._workspace.definition.received_from
```
iakovn, line 470: I suggest adding a property for this:
```
@property
def metadata(self) -> ReceivedFrom:
   """ Reference metadata for the published workspace """
   metadata = self._workspace.definition.received_from
   assert metadata # assuming exception from constructor
   return metadata
```

modelon/impact/client/entities/workspace.py
abhilash-kumar-nair: workspace.rename('renamed workspace')

modelon/impact/client/entities/workspace.py
abhilash-kumar-nair: If we are raising an exception in the constructor, would this be required? Instead if we let the user get an exception that would be more informative for them right? For example if I have code like workspace.recieved_from.get_workspace() I would get an exception like "None" type has no attribute "get_workspace" which I feel is less intuitive to non-programers.

iakovn: agreed

abhilash-kumar-nair:
@iakovn Thanks for the sharp review! Could you please take another look? :)

iakovn (approved):
LGTM ✈️

PR #270, opened 2024-04-09
feat: added support for fetching experiment definition given a experiment result

abhilash-kumar-nair:
This PR contains a first cut implementation of a utility method to create experiment definition based on existing experiment result.

Future: Add support to do this for case result as well.

### DIFF
--- a/modelon/impact/client/entities/experiment.py
+++ b/modelon/impact/client/entities/experiment.py
@@ -7,8 +7,27 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
 from modelon.impact.client import exceptions
 from modelon.impact.client.entities.asserts import assert_variable_in_result
 from modelon.impact.client.entities.case import Case
+from modelon.impact.client.entities.custom_function import CustomFunction
+from modelon.impact.client.entities.external_result import ExternalResult
 from modelon.impact.client.entities.interfaces.experiment import ExperimentInterface
+from modelon.impact.client.entities.model import (
+    Model,
+    SimpleModelicaExperimentDefinition,
+)
+from modelon.impact.client.entities.model_executable import (
+    ModelExecutable,
+    SimpleFMUExperimentDefinition,
+)
 from modelon.impact.client.entities.status import ExperimentStatus
+from modelon.impact.client.experiment_definition.expansion import (
+    ExpansionAlgorithm,
+    FullFactorial,
+    LatinHypercube,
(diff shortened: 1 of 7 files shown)


modelon/impact/client/entities/experiment.py
```diff
@@ -572,3 +597,139 @@ def get_solver_options(self) -> SolverOptions:
...
+    ) -> Optional[CaseOrExperimentOrExternalResult]:
+        if "initializeFrom" in modifiers:
+            exp_id = modifiers["initializeFrom"]
+            return Experiment(
+                workspace_id=self._workspace_id, exp_id=exp_id, service=self._sal
+            )
+        elif "initializeFromCase" in modifiers:
+            case_id = modifiers["initializeFromCase"]["caseId"]
+            exp_id = modifiers["initializeFromCase"]["experimentId"]
+            return Case(case_id, self._workspace_id, exp_id, self._sal, info={})
```
iakovn, line 612: Similar to above - I think you need remote call to retrieve this case by ID from server so that even the "info" is included and also implicitly check that the reference is valid.

modelon/impact/client/entities/experiment.py
```diff
@@ -572,3 +597,139 @@ def get_solver_options(self) -> SolverOptions:
...
         analysis = self._get_info()["experiment"]["base"]["analysis"]
         return SolverOptions(analysis.get("solverOptions", {}), self.custom_function)
+
+    def _get_initialize_from(
+        self, modifiers: Dict[str, Any]
+    ) -> Optional[CaseOrExperimentOrExternalResult]:
+        if "initializeFrom" in modifiers:
+            exp_id = modifiers["initializeFrom"]
+            return Experiment(
+                workspace_id=self._workspace_id, exp_id=exp_id, service=self._sal
```
iakovn, line 607: I think you need remote call to retrieve this experiment by ID from server so that even the "info" is included and also implicitly check that the reference is valid.

iakovn (changes requested):
I like the approach! 🍎 

Some proposals on docstrings.

I think initialize from needs some more thinking and  test coverage.

abhilash-kumar-nair:
Thanks @iakovn  for the review! I have now made the suggested changes. Jenkins seems to error out because of some strange errors.. I can run the tests successfully locally.

iakovn (approved):
Nice work! Let's get this in! 🥇

PR #274, opened 2024-04-17
fix: minor correction to docs in FMU based workflow section

abhilash-kumar-nair:


docs/source/modelon.impact.client.experiment_definition.rst
```diff
@@ -52,6 +52,14 @@ modelon.impact.client.experiment\_definition.model\_based module
...
    :show-inheritance:
 
+modelon.impact.client.experiment\_definition.modifiers module
+-------------------------------------------------------------
+
+.. automodule:: modelon.impact.client.experiment_definition.modifiers
+   :members:
+   :undoc-members:
+   :show-inheritance:
+
```
iakovn, line 62: I don't think we should expose this module in public docs yet. When we do we should document the classes/methods.

iakovn (changes requested):
Minor corrections are fine 💯 but please don't include the new module ⛔

docs/source/modelon.impact.client.experiment_definition.rst
abhilash-kumar-nair: Good catch! :100:

abhilash-kumar-nair: Thanks for the review @iakovn ! I removed the docs addition. Could you please take another look?

iakovn (approved):
LGTM! 👍

PR #275, opened 2024-04-17
Support for fetching experiment definitions from case result

abhilash-kumar-nair:
**Jira** - https://modelonproducts.atlassian.net/browse/FLOW-458


iakovn (approved):
This looks good to me. 👍 

We should consider as separate PR if we really want `create_experiment_definition_from_XXX` functions as public interface. It seems that:
```
workspace.get_experiment(eid).get_definition()
workspace.get_experiment(eid).get_case(cid).get_definition()
```
are pretty convenient and might be sufficient.

PR #277, opened 2024-04-25
feat: added support for new custom function parameter data types

abhilash-kumar-nair:
Jira - https://modelonproducts.atlassian.net/browse/FLOW-483
This PR adds support for CaseResult and ExperimentResult datatypes in python client

### DIFF
--- a/modelon/impact/client/client.py
+++ b/modelon/impact/client/client.py
@@ -12,7 +12,11 @@ import modelon.impact.client.jupyterhub
 import modelon.impact.client.sal.exceptions
 import modelon.impact.client.sal.service
 from modelon.impact.client import exceptions
-from modelon.impact.client.configuration import get_client_interactive, get_client_url
+from modelon.impact.client.configuration import (
+    Experimental,
+    get_client_interactive,
+    get_client_url,
+)
 from modelon.impact.client.credential_manager import CredentialManager
 from modelon.impact.client.entities.case import Case
 from modelon.impact.client.entities.experiment import Experiment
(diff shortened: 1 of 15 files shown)


iakovn (commented):
Nice start!
I'm wondering if we need to reconsider update to:
```
class _Parameter:
 @property
    def value(self) -> Any:
        return self._value
class CustomFunction:
     def parameter_values(self) -> Dict[str, Any]:
```
It seems right now if you set those for ExperimentResult/CaseResult type you need to use Experiment/Case object but when reading back it returns just strings. 

Do you have some examples in mind on how it will work when recreating from previously run experiment/case?

abhilash-kumar-nair:
> Nice start! I'm wondering if we need to reconsider update to:
> 
> ```
> class _Parameter:
>  @property
>     def value(self) -> Any:
>         return self._value
> class CustomFunction:
>      def parameter_values(self) -> Dict[str, Any]:
> ```
> 
> It seems right now if you set those for ExperimentResult/CaseResult type you need to use Experiment/Case object but when reading back it returns just strings.
> 
> Do you have some examples in mind on how it will work when recreating from previously run experiment/case?

Good point! I have created a new class that inherits the built in dict with an additional method that does the conversion to raw dict(that is to be sent to the back-end). Does this make sense?

iakovn (changes requested):
This is fine in one direction. I think we need the reverse as well. 🔙 

Consider:
```
definition = case.get_definition()
cf = definition.custom_function
init_from_case = cf.initialize_from_case
pv = cf.parameter_values
baseline_case = pv['baseline']
```
I expect both `init_from_case` and  `baseline_case` to be either `Case` entity or some new `CaseReference` objects (or None).

I would consider naming to/from protocol somewhere.

modelon/impact/client/entities/case.py
```diff
@@ -789,16 +790,44 @@ def execute(self, sync_case_changes: bool = True) -> CaseOperation:
...
+        custom_function_params_override: Dict[str, Any],
+    ) -> None:
+        for param in custom_function._param_by_name.values():
+            param_override = custom_function_params_override.get(param.name)
+            if param_override:
+                if param.type == "ExperimentResult":
+                    exp_info = self._sal.workspace.experiment_get(
+                        self._workspace_id, param_override
+                    )
+                    custom_function_params_override[param.name] = ExperimentReference(
```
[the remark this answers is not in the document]
abhilash-kumar-nair: Yes it was the circular dependencies. I have now updated the code to return references instead and a utility method to create entities from references.

modelon/impact/client/client.py
```diff
@@ -741,6 +741,22 @@ def get_me(self) -> User:
     def create_entity_from_reference(
```
iakovn, line 741: I suggest for user facing flow two functions marked "experimental":
```
def get_experiment_by_reference(...) -> Experiment
def get_case_by_reference(...) -> Case
```

iakovn (changes requested):
Let's make two experimental functions and merge. I'd like to play in production before finalizing

modelon/impact/client/client.py
abhilash-kumar-nair: Makes sense! Done!

iakovn (approved):
Let's merge! ✈️

PR #278, opened 2024-04-28
Base fixture

abhilash-kumar-nair:
This PR contains some fixes mainly aimed at being able to run/reuse the test suite for integration tests as well

### DIFF
--- a/modelon/impact/client/sal/request.py
+++ b/modelon/impact/client/sal/request.py
@@ -38,26 +38,42 @@ class Request:
         self.headers = headers
 
     def execute(self, check_return: bool = True) -> Any:
+        session_cookies = self.context.session.cookies
+        if "localhost" in self.url:
+            access_token = session_cookies.get("access_token")
+            session_cookies = {"access_token": access_token}
         try:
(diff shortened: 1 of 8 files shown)


modelon/impact/client/sal/request.py
```diff
@@ -38,26 +38,41 @@ def __init__(
         self.headers = headers
 
     def execute(self, check_return: bool = True) -> Any:
+        cookies = self.context.session.cookies
+        access_token = cookies.get("access_token")
+        session_cookies = {"access_token": access_token}
```
iakovn, line 43: This is only needed and should only be done for "localhost" otherwise we should just keep the "session.cookies"

iakovn (changes requested):
Mostly fine. Please restrict cookie mods to the case where we really need that and not otherwise.

modelon/impact/client/sal/request.py
abhilash-kumar-nair: Good point! Fixed it! :)

iakovn (approved):
LGTM!

PR #280, opened 2024-05-03
feat: add convinience label property to the Experiment entity

abhilash-kumar-nair:
This PR adds a convenience label property to the Experiment entity


iakovn (approved):
👍

PR #286, opened 2024-06-06
feat: added support for VariableNames data type for custom function

abhilash-kumar-nair:
This PR adds support for VariableNames data type in Custom Function parameters


iakovn (approved):
LGTM!

PR #289, opened 2024-06-19
feat: added support for result import to case behind experimental flag

abhilash-kumar-nair:
**Jira** - https://modelonproducts.atlassian.net/browse/FLOW-523

**Code to test**

```
import os
os.environ["IMPACT_PYTHON_CLIENT_EXPERIMENTAL"] = "1"

client = Client(url="http://localhost:8888/impact", interactive=True)
workspace = client.get_workspace("test")

# Choose analysis type

# Get model class

# Execute experiment
experiment_definition = model.new_experiment_definition(dynamic)
case_1 = exp.get_case('case_1')
result = case_1.import_result('/home/akn/Downloads/Result.csv', overwrite=True).wait()
print(result.keys())
```

### DIFF
--- a/modelon/impact/client/entities/case.py
+++ b/modelon/impact/client/entities/case.py
@@ -21,6 +21,7 @@ from modelon.impact.client.entities.result import Result
 from modelon.impact.client.entities.status import CaseStatus
 from modelon.impact.client.operations.base import BaseOperation
 from modelon.impact.client.operations.case import CaseOperation
+from modelon.impact.client.operations.case_result import CaseResultImportOperation
 from modelon.impact.client.operations.custom_artifact import (
     CustomArtifactImportOperation,
 )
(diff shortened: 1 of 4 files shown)


maraberg:
I think the script looks a bit weird - aren't we supposed to use "empty" cf in order for result upload to be valid?

iakovn:
> I think the script looks a bit weird - aren't we supposed to use "empty" cf in order for result upload to be valid?

We moved back to using (any) actual custom functions and not empty. If your only intention is to load results you can create a custom function "load_results" in the library.

maraberg:
Minor thing: Do we do any checks on the backend on the csv data types, I managed to upload a csv - file with semi-colons as dividers: this was the result:
image

maraberg:
For multi-case simulation it seems like the result.keys() are not correct:
image

maraberg:
Same with trajectory object:

image

maraberg:
[sandbox.zip](https://github.com/user-attachments/files/15901651/sandbox.zip)

Zip with resources used for tests

maraberg:
I have not looked at the code changes - let me know if you think this is valuable, I don't think so myself :)

abhilash-kumar-nair:
> Minor thing: Do we do any checks on the backend on the csv data types, I managed to upload a csv - file with semi-colons as dividers: this was the result: image

Good catch. I will add checks for delimiter and raise appropriate errors while validating such results

abhilash-kumar-nair:
> ny checks on the backend on the csv data types, I managed to upload a csv - file with semi-colons as dividers: this was the result:

We would need another api to fetch CaseResult variables. Currently we only have one that operates on the experiment level. This was your concern right?

The current GET /variables api seems to correctly capture the variables in your simulated result file and the overwritten result file.

abhilash-kumar-nair:
> > Minor thing: Do we do any checks on the backend on the csv data types, I managed to upload a csv - file with semi-colons as dividers: this was the result: image
> 
> Good catch. I will add checks for delimiter and raise appropriate errors while validating such results

Thanks a lot for the user feedback @maraberg ! I will work on making the changes you suggested and get back to you for another round of testing :)

PR #290, opened 2024-06-20
Single case vars

abhilash-kumar-nair:
Jira - https://modelonproducts.atlassian.net/browse/FLOW-558

This MR adds support for fetching result variables for a single case. This api addresses the concerns raised by @maraberg  in https://github.com/modelon-community/impact-client-python/pull/289 where result variables for a multi-execution case didnt contain case specific result variables.

### DIFF
--- a/modelon/impact/client/client.py
+++ b/modelon/impact/client/client.py
@@ -222,7 +222,7 @@ class Client:
 
     """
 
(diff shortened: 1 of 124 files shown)


abhilash-kumar-nair:
> I'm not able to test conveniently without CIMP-MR build, however from what I understand from the code changes it looks good. So previously we always returned the variables from the experiment, even for the cases. This looks like it is fixed here :)

Yes thats correct! :)

PR #292, opened 2024-06-24
Result import

abhilash-kumar-nair:
This MR adds API support for importing result files for a case and also list result variables for a case.

### DIFF
--- a/modelon/impact/client/client.py
+++ b/modelon/impact/client/client.py
@@ -222,7 +222,7 @@ class Client:
 
     """
 
-    _SUPPORTED_VERSION_RANGE = ">=4.11.0,<5.0.0"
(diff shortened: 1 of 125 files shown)


maraberg:
Not 100% what to review here :) Anyway, I don't see any added unit tests with the new functionality?

abhilash-kumar-nair:
> Not 100% what to review here :) Anyway, I don't see any added unit tests with the new functionality?



> Not 100% what to review here :) Anyway, I don't see any added unit tests with the new functionality?

The PR you tested on diraction contained these changes so we should be good. I will have another MR to remove experimental flag for a couple of new API we have added and also would add tests there. We have another deploy planned during the second week of July when this can go to public cloud.

maraberg (approved):

iakovn (approved):
LGTM! 👍

PR #293, opened 2024-06-25
chore: update compatibility matrix in docs

abhilash-kumar-nair:


abhilash-kumar-nair:
@iakovn , I have updated the table for MI versions as users have run into this when using production cloud. I will update the table with open-api versions in another PR

iakovn (approved):
👍

PR #297, opened 2024-07-11
fix: pass user defined context to unauthenticated HTTPClient

abhilash-kumar-nair:
This PR fixes such that a user defined context is passed to the unauthenticated HttpClient as well.


iakovn (approved):
LGTM! 👍

PR #303, opened 2024-07-29
fix: updated the AccessSettings object with updated protocol for workspace publish

abhilash-kumar-nair:
[Jira - https://modelonproducts.atlassian.net/browse/FLOW-609](https://modelonproducts.atlassian.net/browse/FLOW-609)


iakovn (approved):
LGTM! 👍

PR #304, opened 2024-07-30
Query as Params

abhilash-kumar-nair:
The requests library supports query strings to be sent for a request in the params arument See https://requests.readthedocs.io/en/latest/user/quickstart/#passing-parameters-in-urls. This is a safer and neater approach than embedding them as raw strings in the url.

### DIFF
--- a/modelon/impact/client/sal/workspace.py
+++ b/modelon/impact/client/sal/workspace.py
@@ -24,10 +24,9 @@ class WorkspaceService:
         self._http_client.delete_json(url)
 
     def workspace_get(self, workspace_id: str, size_info: bool) -> Dict[str, Any]:
-        url = (
-            self._base_uri / f"api/workspaces/{workspace_id}?sizeInfo={size_info}"
-        ).resolve()
-        return self._http_client.get_json(url)
+        query = {"sizeInfo": size_info}
+        url = (self._base_uri / f"api/workspaces/{workspace_id}").resolve()
+        return self._http_client.get_json(url, params=query)
 
     def workspaces_get(
         self,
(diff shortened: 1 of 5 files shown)


modelon/impact/client/sal/workspace.py
```diff
@@ -226,24 +209,20 @@ def get_published_workspaces(
...
-            query["workspaceName"] = name
-        if has_data:
-            query["hasData"] = str(has_data)
+        query = {
+            "workspaceName": name,
+            "hasData": has_data,
+            "ownerUsername": owner_username,
+            "type": type,
+            "groupName": group_name,
+        }
```
iakovn, line 218: Why did you skip all the conditions? It seems unnecessary to pass, e.g.,  `workspaceName=None` all the way to the server?

abhilash-kumar-nair: I felt this makes the python client api consistent, if the defaults in server are updated for some reason? The server would anyway parse the query strings and set defaults if they are not set here.  I agree for workspaceName probably it doesnt make sense to send in.

iakovn: I agree with this argument for boolean fields (like  `has_data`) and eventually pagination control (first/max) but I feel it is a bit weird for optional strings like workspaceName, ownerUsername and groupName. Are those getting serialized like `workspaceName=null` when passed to requests?

abhilash-kumar-nair: Just read the docs, seems like requests would not append it to the URL if its **NONE** its seems as per docs - https://docs.python-requests.org/en/latest/user/quickstart/#passing-parameters-in-urls

For empty string they would just be added like https://blahblah.com/api?workspaceName=&ownerUsername=

iakovn: Thanks for double-checking! Then I'm fine with the update.

iakovn:
LGTM! 👍

iakovn (approved):

PR #311, opened 2024-09-06
fix: ensure that client can always be imported with other modelon packages in all cases

efredriksson-modelon:
This MR is to unify how other 'modelon-impact-*' packages handles the top level init files. Without this it is for example not possible to source install another 'modelon-impact-*' package if the client package is installed (as only modelon package or installed last).


iakovn (approved):
👍

PR #313, opened 2024-09-10
chore: update compile FMU tests to be more deterministic

abhilash-kumar-nair:
The integration tests were failing due to change in OCT as reported here - https://modelonproducts.atlassian.net/browse/INS-841. (See comments) which added a new internal parameter in the settable parameters.


iakovn (approved):
👍

PR #315, opened 2024-09-30
fix: accept any artifact content type

iakovn:
See FLOW-652:
Artifact download end-point can return different content types depending on the file extension. Client should accept any.


abhilash-kumar-nair (approved):
Good catch! Thanks for fixing this. The changes look good! I have updated the test and regenerated the cassette now.

PR #316, opened 2024-10-01
fix: passing values as positional argument in Choices operator's from_dict method instead of keyword argument

abhilash-kumar-nair:
This PR fixes - INS-898.


iakovn (approved):
LGTM! 👍

PR #318, opened 2024-10-03
feat: added support to send modelica resource uri's as modifiers

abhilash-kumar-nair:
Story - CAP-909


iakovn (commented):
I believe the implementation is sound. I would like to wait a bit with and test with the compiler and backend versions that are due in a couple of weeks.

PR #320, opened 2024-11-21
chore(docs): added quickstart for setting sim opts for fmu based workflow

abhilash-kumar-nair:
This PR improves the quickstart for FMU based experiment with snippet to set simulation and solver options.


iakovn (approved):
👍

PR #323, opened 2025-02-20
chore: updates forJH 5.2.1/MI 4.24.0

iakovn:
- Updated build system and regenerated cassettes.
- Added scrubbing of user id
- Improved "get_me" test robustness when running in different modes
- Removed refs to "c_compiler" options that will be excluded from defaults


abhilash-kumar-nair (approved):
Thanks for cleaning this up! I have tested the regeneration against prod cloud and it works well! :watermelon:

PR #325, opened 2025-03-10
feat: add method for fetching experiment size in Experiment entity

abhilash-kumar-nair:
This PR fixes - https://modelonproducts.atlassian.net/browse/WAMS-14179

### DIFF
--- a/modelon/impact/client/entities/experiment.py
+++ b/modelon/impact/client/entities/experiment.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import enum
 import logging
 import warnings
+from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
 
 from modelon.impact.client.entities.asserts import assert_variable_in_result
@@ -98,6 +99,11 @@ class _Workflow(enum.Enum):
 
     FMU_BASED = "FMU_BASED"
     CLASS_BASED = "CLASS_BASED"
+
+
+@dataclass
+class ExperimentSizeInfo:
+    total: int
 
 
 class ExperimentRunInfo:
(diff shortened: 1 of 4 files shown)


modelon/impact/client/entities/experiment.py
```diff
@@ -214,6 +224,13 @@ def _get_info(self, cached: bool = True) -> Dict[str, Any]:
 
         return self._info
 
+    def get_size_info(self) -> ExperimentSizeInfo:
+        """The experiment size information."""
+        run_info = self._sal.workspace.experiment_get(
+            self._workspace_id, self._exp_id, size_info=True
+        )["run_info"]
+        return ExperimentSizeInfo(run_info.get("sizeInfo", {}))
```
iakovn, line 232: Why are you checking for empty dict here? It seems more like an exception of some kind?

modelon/impact/client/entities/experiment.py
```diff
@@ -100,6 +100,16 @@ class _Workflow(enum.Enum):
     CLASS_BASED = "CLASS_BASED"
 
 
+class ExperimentSizeInfo:
+    def __init__(self, data: Dict[str, Any]):
+        self._data = data
+
+    @property
+    def total(self) -> Optional[int]:
```
iakovn, line 108: Why do you put Optional[int]?

modelon/impact/client/entities/experiment.py
```diff
@@ -100,6 +100,16 @@ class _Workflow(enum.Enum):
     CLASS_BASED = "CLASS_BASED"
 
 
+class ExperimentSizeInfo:
+    def __init__(self, data: Dict[str, Any]):
```
iakovn, line 104: I'd suggest parsing out the integer directly and raising exception if the expected key is not present. It feels unnecessary to delay that until the "total" call and have Optional[int]

iakovn (commented):
Thanks for taking this up so quickly! 🥕 

Is there a good reason to have sizeInfo as optional when requested? Why not handling missing size info as an error?

modelon/impact/client/entities/experiment.py
abhilash-kumar-nair: I was thinking a bit about extensibility. If the "sizeInfo" object contains more data in the future... But I agree on throwing an exception is a better approach here. I have also made ExperimentSizeInfo to a dataclass now.

abhilash-kumar-nair:
> Thanks for taking this up so quickly! 🥕
> 
> Is there a good reason to have sizeInfo as optional when requested? Why not handling missing size info as an error?

I agree raising an exception make more sense. Fixed it now.

iakovn (approved):
LGTM! 🏃‍♂️

PR #326, opened 2025-04-02
chore: remove usage of c_compiler opts from vcr tests

abhilash-kumar-nair:


iakovn (commented):
It'd be nice to understand why id's are changing as that is the largest part of the changeset.
Do we really need to regenerate cassetes?

iakovn (approved):
LGTM! 👍

PR #327, opened 2025-04-15
feat: add user agent that is specific for the client

efredriksson-modelon:
In order for the python-client to uniquely identify itself against the server, we need to add a specific header in order to separate it from other consumers.


abhilash-kumar-nair:
Looks good! :carrot:

abhilash-kumar-nair (approved):

PR #332, opened 2025-08-18
fix: corrected the regex for file_name match for file_name property

abhilash-kumar-nair:
This PR fixes https://modelonproducts.atlassian.net/browse/INS-1488. 
@efredriksson-modelon ,  this fix may likely cause older MI versions to return an empty string. Should we do this in a backward compatible way?

### DIFF
--- a/modelon/impact/client/sal/response.py
+++ b/modelon/impact/client/sal/response.py
@@ -165,7 +165,13 @@ class FileResponse(Response):
     @property
     def file_name(self) -> str:
         d = self.headers["content-disposition"]
-        return re.findall("filename=(.+)", d)[0].strip('"')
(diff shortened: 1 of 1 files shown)


abhilash-kumar-nair:
> @abhilash-kumar-nair I think it makes sense to try checking the new utf-8 first and check in the old way if it returns empty.
> 
> FYI, there seems to be some issue with the regexp in the tests.

@efredriksson-modelon Sure! Fixed it! :) Looks like we still need to support the old way anyways, since resp.downloadable_as = str(file_name) used in some APIs would put the filename in content-disposition headers in the old way.

efredriksson-modelon (approved):
Thanks :watermelon:

PR #339, opened 2026-05-20
Support for Introspection API's

abhilash-kumar-nair:
This PR adds support for introspecting the Model source, Model parameters, Top level packages and recursively look into their subpackages/models.

```
print("Top level packages")
with ws.new_modeling_session() as session:
    for package in session.get_top_level_packages():
        print(f"  {package.name}")
        if package.name == "Unnamed":
            print("  Sub-packages:", package.get_subpackages())
            print("  Sub-models:", package.get_models())
    model = session.get_model("Unnamed.test")
    print("Modelica source code")
    print(model.get_source())
    print("Model parameters")
    print(model.get_parameters())
````

### DIFF
--- a/modelon/impact/client/entities/model.py
+++ b/modelon/impact/client/entities/model.py
@@ -2,8 +2,10 @@ from __future__ import annotations
 
 import logging
 import os
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Union
 
+from modelon.impact.client.configuration import Experimental
 from modelon.impact.client.entities.interfaces.model import ModelInterface
 from modelon.impact.client.entities.model_executable import ModelExecutable
 from modelon.impact.client.entities.project import Project
@@ -30,11 +32,18 @@ if TYPE_CHECKING:
     from modelon.impact.client.entities.experiment import Experiment
     from modelon.impact.client.entities.external_result import ExternalResult
     from modelon.impact.client.operations.base import BaseOperation
+    from modelon.impact.client.sal.modeling import ModelingService
     from modelon.impact.client.sal.service import Service
 
     CaseOrExperimentOrExternalResult = Union[Case, Experiment, ExternalResult]
 
 logger = logging.getLogger(__name__)
+
+
+@dataclass
+class ModelParameter:
+    name: str
+
 
 CompilationOperations = Union[ModelExecutableOperation, CachedModelExecutableOperation]
 
@@ -78,12 +87,18 @@ class Model(ModelInterface):
     """Class containing Model functionalities."""
 
     def __init__(
-        self, class_name: str, workspace_id: str, project_id: str, service: Service
+        self,
+        class_name: str,
+        workspace_id: str,
+        project_id: str,
+        service: Service,
+        modeling_sal_getter: Optional[Callable[[], "ModelingService"]] = None,
     ):
         self._class_name = class_name
         self._workspace_id = workspace_id
         self._project_id = project_id
         self._sal = service
+        self._modeling_sal_getter = modeling_sal_getter
 
     def __repr__(self) -> str:
         return f"Model name '{self._class_name}'"
@@ -398,6 +413,48 @@ class Model(ModelInterface):
             self._sal,
             Model.from_operation,
         )
+
+    def _require_modeling_session(
+        self, method_name: str
+    ) -> Callable[[], "ModelingService"]:
+        if self._modeling_sal_getter is None:
+            raise RuntimeError(
+                "This Model instance has no modeling session. "
+                f"Obtain the model via modeling_session.get_model() to enable "
+                f"{method_name}()."
+            )
+        return self._modeling_sal_getter
+
+    @Experimental
+    def get_parameters(self) -> List[ModelParameter]:
+        """Returns the parameters for this model's class.
+
+        Example::
+
+            with workspace.new_modeling_session() as session:
+                model = session.get_model("LibA.Model")
+                parameters = model.get_parameters()
+
+        """
+        modeling_sal = self._require_modeling_session("get_parameters")
+        return [
+            ModelParameter(name=name)
+            for name in modeling_sal().get_model_parameters(self._class_name)
+        ]
+
+    @Experimental
+    def get_source(self) -> str:
+        """Returns the Modelica source code for this model's class.
+
+        Example::
+
+            with workspace.new_modeling_session() as session:
+                model = session.get_model("LibA.Model")
+                source = model.get_source()
+
+        """
+        modeling_sal = self._require_modeling_session("get_source")
+        return modeling_sal().get_model_source(self._class_name)
 
     @classmethod
     def from_operation(cls, operation: BaseOperation[Model], **kwargs: Any) -> Model:
--- a/modelon/impact/client/entities/modeling.py
+++ b/modelon/impact/client/entities/modeling.py
@@ -1,0 +1,76 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, List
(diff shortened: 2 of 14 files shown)

### CODE REVIEW

File: modelon/impact/client/sal/modeling.py [hunk 1]
```diff
@@ -0,0 +1,44 @@
...
+from modelon.impact.client.sal.uri import URI
+from modelon.impact.client.sal.ws import SyncWebSocketClient
+
+
+class ModelingService:
+    def __init__(
+        self, uri: URI, http_client: HTTPClient, ws_client: SyncWebSocketClient
+    ):
+        self._base_uri = uri
+        self._http_client = http_client
```
efredriksson-modelon: Not used and can be removed?

pyproject.toml
```diff
@@ -14,18 +14,19 @@ authors = [
...
-    "Programming Language :: Python :: 3.8"
+    "Programming Language :: Python :: 3.9"
 ]
 include = [
 ]
 keywords = ["impact", "client", "API"]
 
 [tool.poetry.dependencies]
-python = "^3.8.0"
+python = ">=3.9.0,<3.12.0"
```
efredriksson-modelon: I think the upper limit should be 4.0. Adding the 3.12 limit can cause a lot of issues for downstream tooling.

modelon/impact/client/entities/workspace.py
```diff
@@ -646,13 +648,39 @@ def __init__(
...
+    def __exit__(self, *args: object) -> None:
+        self.close()
+
+    def _get_modeling_session(self) -> ModelingSession:
+        if self._modeling_session is None:
+            modeling_service = self._sal.start_modeling_session(workspace_id=self.id)
+            self._modeling_session = ModelingSession(
+                modeling_service.id, self.id, modeling_service, self._sal
+            )
+        return self._modeling_session
```
efredriksson-modelon: This machinery adds a lot of hidden state and makes it easy for a user to setup things so things becomes unstable. I think that we should drop it and accept that the get_model on workspace does not have full functionality. This seems pretty minor given the reduced complexity.

modelon/impact/client/entities/model.py
```diff
@@ -399,6 +407,40 @@ def import_fmu(
             Model.from_operation,
         )
 
+    @Experimental
+    def get_parameters(self) -> List[str]:
```
efredriksson-modelon: Is this returning the parameter names? If we want to add things in the future (start value/nominals/...) it might we worth return `list[ModelParameter]` where `ModelParameter` have the attribute `name`.

modelon/impact/client/entities/modeling.py
```diff
@@ -0,0 +1,41 @@
...
+        return f"Modeling session with ID '{self._session_id}'"
+
+    def close(self) -> None:
+        return self._modeling_sal.close_session()
+
+    def close_nowait(self) -> None:
+        return self._modeling_sal.close_session_nowait()
+
+    @Experimental
+    def get_model_parmeters(self, modelica_path: str) -> List[str]:
```
efredriksson-modelon: Same here on `list[str]`. But should we even have these methods? I would have expected something like:
```
model_session.get_model(class_path).get_parameters()
model_session.get_model(class_path).get_source()
```
so we keep the API clean and give the user one way to do this.

modelon/impact/client/entities/package.py
```diff
@@ -0,0 +1,80 @@
...
+        """Returns the libraries the package has dependencies on."""
+        return self._info()["uses"]
+
+    def get_sub_models(self) -> List[Model]:
+        return [
+            Model(class_path, self._workspace_id, self._project_id, self._sal)
+            for class_path in self._modeling_sal.get_sub_models(self._class_path)
+        ]
+
+    def get_sub_packages(self) -> List["Package"]:
```
efredriksson-modelon: sub_packages -> subpackages seems to be the terminology used: https://specification.modelica.org/master/packages.html

modelon/impact/client/entities/package.py
```diff
@@ -0,0 +1,80 @@
...
+    @property
+    def is_enabled(self) -> bool:
+        """Returns True if the package is enabled."""
+        return self._info()["enabled"]
+
+    def get_uses(self) -> Dict[str, str]:
+        """Returns the libraries the package has dependencies on."""
+        return self._info()["uses"]
+
+    def get_sub_models(self) -> List[Model]:
```
efredriksson-modelon: I don't think sub_models are a thing. I think this should just be called  `get_models`.

modelon/impact/client/entities/workspace.py
```diff
@@ -678,6 +706,20 @@ def name(self) -> str:
         """Workspace name."""
         return self.definition.name
 
+    @Experimental
+    @contextmanager
+    def start_modeling_session(self) -> Generator["ModelingSession", None, None]:
```
efredriksson-modelon: `start_modeling_session` -> `new_modeling_session` might read a little nicer and match our other naming, but feel free to ignore this.

modelon/impact/client/sal/service.py
```diff
@@ -57,6 +61,12 @@ def __init__(self, uri: URI, api_key: str, context: Optional[Context] = None):
             if is_jupyterhub_url(uri, context)
             else uri
         )
+        _is_local_dev = (
+            os.environ.get("IMPACT_PYTHON_CLIENT_DEV", "false").lower() == "true"
+        )
+        scheme = "ws" if _is_local_dev else "wss"
```
efredriksson-modelon, line 67: Do we want to keep this in?

modelon/impact/client/sal/ws.py
```diff
@@ -0,0 +1,128 @@
...
+
+    async def connect(self) -> None:
+        headers = {"User-Agent": "impact-python-client"}
+        if self._api_key:
+            headers["impact-api-key"] = self._api_key
+
+        self._conn = await websockets.connect(
+            self._uri,
+            additional_headers=headers,
+        )  # type: ignore
```
efredriksson-modelon: I might be missing something here but you are creating an async connection and then you want to wrap it in a sync API? I have not tested it, but could we not use `websockets.sync.client.connect` directly?

modelon/impact/client/entities/model.py
```diff
@@ -399,6 +407,40 @@ def import_fmu(
...
+        Example::
+
+            model = workspace.get_model("LibA.Model")
+            parameters = model.get_parameters()
+
+        """
+        if self._modeling_sal_getter is None:
+            raise RuntimeError(
+                "This Model instance has no modeling session. "
+                "Obtain the model via workspace.get_model() to enable get_parameters()."
```
efredriksson-modelon: Obtain the model via modeling_session.get_model, see workspace.new_model_session is what this would say.

efredriksson-modelon (changes requested):
Looks really cool 🧇 

I think the main thing is to have `get_model` on the `modeling_session` object and accept that `workspace.get_model().get_source()` will not work. Otherwise there are too many ways you can create issues and this also streamlines the design significantly.

I also wonder about the async vs sync websocket client connection, which might be good to take a look at.

modelon/impact/client/sal/service.py
abhilash-kumar-nair: Yes for local testing with cimp from outside JL :)

abhilash-kumar-nair:
> Looks really cool 🧇
> 
> I think the main thing is to have `get_model` on the `modeling_session` object and accept that `workspace.get_model().get_source()` will not work. Otherwise there are too many ways you can create issues and this also streamlines the design significantly.
> 
> I also wonder about the async vs sync websocket client connection, which might be good to take a look at.

Thanks @efredriksson-modelon ! I have now made the suggeted changes. Could you please take another peek?

efredriksson-modelon (approved):
🍉

PR #342, opened 2026-06-08
feat: add api to fetch the available experiment definitions for a model

abhilash-kumar-nair:
This PR adds an experimental api to fetch the experiment definitions for a model

**Test script**

```
import os
os.environ["IMPACT_PYTHON_CLIENT_EXPERIMENTAL"]="true"

client=Client(url="url",interactive=True)
ws=client.get_workspace("tes")
model=ws.get_model("Unnamed.PID_Controller")
exp_defs = model.get_experiment_definitions()
for exp_def in exp_defs:
    print(exp_def.id)

```

### DIFF
--- a/modelon/impact/client/entities/model.py
+++ b/modelon/impact/client/entities/model.py
@@ -6,6 +6,7 @@ from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Union
 
 from modelon.impact.client.configuration import Experimental
+from modelon.impact.client.entities.custom_function import CustomFunction
 from modelon.impact.client.entities.interfaces.model import ModelInterface
 from modelon.impact.client.entities.model_executable import ModelExecutable
 from modelon.impact.client.entities.project import Project
@@ -28,7 +29,6 @@ from modelon.impact.client.options import (
 
 if TYPE_CHECKING:
     from modelon.impact.client.entities.case import Case
-    from modelon.impact.client.entities.custom_function import CustomFunction
     from modelon.impact.client.entities.experiment import Experiment
     from modelon.impact.client.entities.external_result import ExternalResult
     from modelon.impact.client.operations.base import BaseOperation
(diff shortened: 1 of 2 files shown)


modelon/impact/client/sal/workspace.py
```diff
@@ -355,6 +355,23 @@ def workspace_lock_preview(
...
+        model_name: str,
+        extends: Optional[List[str]] = None,
+    ) -> Dict[str, Any]:
+        url = (
+            self._base_uri / f"api/workspaces/{workspace_id}/models/{model_name}/"
+            "experiment-definitions"
+        ).resolve()
+        params: Dict[str, Any] = {}
+        if extends:
+            params["extends"] = extends
```
efredriksson-modelon: Could be a bit more compact: `params = {"extends": extends} if extends else {}`

modelon/impact/client/entities/model.py
```diff
@@ -224,6 +237,72 @@ def compile(
...
+            custom_function_params = {
+                param["name"]: param["value"]
+                for param in analysis.get("parameters", [])
+            }
+            custom_function = CustomFunction(
+                self._workspace_id,
+                custom_function_meta["name"],
+                custom_function_meta["parameters"],
+                self._sal,
+            ).with_parameters(**custom_function_params)
```
efredriksson-modelon, line 267: We definitely need to refactor out some common code for converting dict to experiment definition. This is the 3rd implementation for creating CustomFunction from dict in the code base. I think similar can be said about other dict -> classes that is done here for the experiment definition.

efredriksson-modelon (changes requested):

modelon/impact/client/entities/model.py
abhilash-kumar-nair: I totally agree. I started refactoring a bit but encountered a few cyclic dependencies. Is it OK if I will create another MR with a cleaned up version as it touches a couple of modules unrelated to this?

efredriksson-modelon: Sure sounds good. I think it can probably be a new module for just doing the mapping so I think that should help with cyclic dependencies?

efredriksson-modelon (approved):

PR #343, opened 2026-06-08
refactor: add factory classmethods for building CustomFunction and SimpleModelicaExperimentDefinition from API response dicts

abhilash-kumar-nair:
This MR refactors the code to have factory classmethods for building CustomFunction and SimpleModelicaExperimentDefinition from API response dicts as discussed in https://github.com/modelon-community/impact-client-python/pull/342#discussion_r3371476200

### DIFF
--- a/modelon/impact/client/entities/experiment.py
+++ b/modelon/impact/client/entities/experiment.py
@@ -8,7 +8,10 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
 
 from modelon.impact.client.entities.asserts import assert_variable_in_result
 from modelon.impact.client.entities.case import Case
-from modelon.impact.client.entities.custom_function import CustomFunction
+from modelon.impact.client.entities.custom_function import (
+    CustomFunction,
+    _build_custom_function,
+)
 from modelon.impact.client.entities.external_result import ExternalResult
 from modelon.impact.client.entities.interfaces.experiment import ExperimentReference
 from modelon.impact.client.entities.model import (
@@ -739,7 +742,7 @@ class Experiment(ExperimentReference):
             self._workspace_id, self.custom_function
         )
         params = {p["name"]: p["value"] for p in analysis.get("parameters", [])}
-        return CustomFunction.from_response_dict(
+        return _build_custom_function(
             self._workspace_id, meta, self._sal
         ).with_parameters(**params)
 
--- a/modelon/impact/client/entities/experiment.py
+++ b/modelon/impact/client/entities/experiment.py
@@ -25,12 +25,6 @@ from modelon.impact.client.entities.model_executable import (
     SimpleFMUExperimentDefinition,
 )
 from modelon.impact.client.entities.status import ExperimentStatus
-from modelon.impact.client.experiment_definition.expansion import (
-    ExpansionAlgorithm,
-    FullFactorial,
-    LatinHypercube,
-    Sobol,
-)
 from modelon.impact.client.experiment_definition.extension import (
     SimpleExperimentExtension,
 )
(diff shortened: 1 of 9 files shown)


modelon/impact/client/experiment_definition/model_based.py
```diff
@@ -157,6 +157,31 @@ def __init__(
         self._extensions: List[SimpleExperimentExtension] = []
         self._expansion: ExpansionAlgorithm = FullFactorial()
 
+    @classmethod
+    def from_experiment_base_dict(
```
efredriksson-modelon: Do we want to have these as public methods on these object that the user is suppose to work with? It kind of pollutes the API, why it would be nicer to have this separate to reduce exposure.

modelon/impact/client/experiment_definition/model_based.py
```diff
@@ -157,6 +157,31 @@ def __init__(
...
+        cls,
+        model: Model,
+        base: Dict[str, Any],
+        custom_function: CustomFunction,
+        initialize_from: Optional[CaseOrExperimentOrExternalResult] = None,
```
efredriksson-modelon: It seems we already have functionality for converting initialize_from from dict to domain. I think it would have been nicer to call that function inside this one to ensure things are consistent.

efredriksson-modelon (commented):
Looks like a very nice cleanup 🧇 

Looking at the code I don't see any reason these methods need to be on the objects themself. It would be nice to keep them as clean as possible as users will handle them in their scripts and tab on them, keeping pollution minimal will help the user here.

abhilash-kumar-nair:
> Looks like a very nice cleanup 🧇
> 
> Looking at the code I don't see any reason these methods need to be on the objects themself. It would be nice to keep them as clean as possible as users will handle them in their scripts and tab on them, keeping pollution minimal will help the user here.

Fair point! I have moved them out now. Also fixed issues with modifiers and expansions.

modelon/impact/client/entities/experiment.py
```diff
@@ -704,11 +678,11 @@ def get_definition(self) -> ValidExperimentDefinitions:
...
-        }
-        definition = definition.with_modifiers(modifiers=modifiers)
+            modifiers = {
+                mod["name"]: get_operator_from_dict(mod)
+                for mod in base["modifiers"]["variables"]
+            }
+            definition = definition.with_modifiers(modifiers=modifiers)
         extensions = self._get_info()["experiment"].get("extensions")
         if extensions:
             sim_exts = []
```
efredriksson-modelon, line 688: It would have been nice to handle `SimpleFMUExperimentDefinition` and `extensions` in `_build_simple_modelica_experiment_definition` but I guess it is OK for now as experiment definitions can't have extensions. However, something it would be nice to fix so we don't run into a bug related to this later.

modelon/impact/client/experiment_definition/model_based.py
```diff
@@ -57,6 +61,47 @@
...
+
+    modelica = base["model"]["modelica"]
+    analysis = base["analysis"]
+    variable_modifiers = {
+        mod["name"]: get_operator_from_dict(mod)
+        for mod in base.get("modifiers", {}).get("variables", [])
+    }
+    expansion_dict = base.get("expansion", {})
+    expansion = expansion_from_dict(
+        expansion_dict.get("algorithm", ""), expansion_dict.get("parameters", {})
```
efredriksson-modelon, line 83: Would have made more sense to send in the entire `expansion_dict` as a single argument here?

efredriksson-modelon (approved):
Nice, let's get this in 🍉 

I think taking another pass would also be nice to ensure future bugs are also covered but that can wait.

abhilash-kumar-nair:
@efredriksson-modelon , Thanks for the review! I will push further refactors in another PR

PR #344, opened 2026-06-12
fix: keep modifiers when fetching experiment definitions for a model

rfatideh:
## Problem

`Model.get_experiment_definitions()` (added in #d4b80c7 / 4.13.0) reconstructs a `SimpleModelicaExperimentDefinition` from each stored definition item but never reads `base.modifiers`. As a result the returned `entry.definition` always has empty variable modifiers, no initialize-from result, and the default expansion algorithm — and `workspace.execute(entry.definition)` serializes `"modifiers": {"variables": []}`, silently running the experiment without any of its saved modifiers.

Example: a definition saved with

```json
"modifiers": {"initializeFrom": "", "variables": [{"name": "driveAngle", "expression": "1.65806278939461138713"}]}
```

executes with `driveAngle` at its model default instead of the saved value.

## Fix

`get_experiment_definitions` now restores the stored modifiers, mirroring what `Experiment.get_definition()` does for executed experiments:

- **Variable modifiers** are populated from `base.modifiers.variables`. Both shapes are supported:
  - experiment-schema operator dictionaries (`{"kind": "value"|"range"|..., ...}`), parsed with the existing `get_operator_from_dict`;
  - textual expressions (`{"name": ..., "expression": "1.65"|"range(0.1,0.5,3)"|...}`) as saved by the Impact UI, parsed with a new `get_operator_from_expression` helper that mirrors the UI's expression-to-modifier conversion (`modifierV2ToV3`).
- **Initialize-from** (`initializeFrom`, `initializeFromCase`, `initializeFromExternalResult`) is resolved to the corresponding entity. The empty string and the UI session sentinel `"latest"` are skipped since they cannot be resolved outside the editor.
- **Expansion algorithm** is restored from `base.expansion` when present.
- **`Choices` with enumeration values**: `Choices(*values, data_type=DataType.ENUMERATION)` previously raised `ValueError` because enumeration literals are plain strings. This also made `Choices.from_dict` (and therefore `Experiment.get_definition()`) fail for experiments using enumeration choices. String values are now accepted when the data type is ENUMERATION.

## Not covered

Saved definition `extensions` are still not reconstructed (out of scope for this fix; the entry-level reconstruction never handled them and the UI does not store them in definitions).

## Testing

- New unit tests for `get_experiment_definitions` covering both modifier shapes, initialize-from resolution, the `""`/`"latest"` guards, and definitions without modifiers (`tests/impact/client/entities/test_model.py`).
- New unit tests for `get_operator_from_expression` (scalars, enumerations, quoted strings, `loadResource`, all operator kinds) and for ENUMERATION `Choices` (`tests/impact/client/test_experiment_definition.py`).
- Full suite: 329 passed; the only failure (`test_get_client_url_default`) is a pre-existing environment-dependent test unrelated to this change.
- Lint (black, flake8, mypy, isort, docformatter, pylint spelling) is clean.

### DIFF
--- a/modelon/impact/client/experiment_definition/operators.py
+++ b/modelon/impact/client/experiment_definition/operators.py
@@ -1,6 +1,7 @@
 """This module contains operators for parametrizing batch runs."""
 from __future__ import annotations
 
+import re
 from abc import abstractmethod
 from dataclasses import dataclass
 from typing import Any, Optional, Union
@@ -109,6 +110,10 @@ class Choices(Operator):
 
             if used_data_type == DataType.REAL and value_type == DataType.INTEGER:
                 pass  # Integer can be assigned to real, do nothing
+            elif (
+                used_data_type == DataType.ENUMERATION and value_type == DataType.STRING
+            ):
+                pass  # Enumeration literals are given as strings, do nothing
             elif used_data_type != value_type:
                 if given_data_type:
                     raise ValueError(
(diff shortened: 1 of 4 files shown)


efredriksson-modelon:
@rfatideh I would suggest you check with @abhilash-kumar-nair when he have time. This is related to https://github.com/modelon-community/impact-client-python/pull/343 and making sure de-serialization code is re-used.

modelon/impact/client/experiment_definition/operators.py
```diff
@@ -313,3 +318,66 @@ def get_operator_from_dict(
...
+    if "loadResource" in expression:
+        return expression
+    if "." in expression:
+        return Enumeration(expression)
+    return expression
+
+
+def get_operator_from_expression(
+    expression: str,
+) -> Union[Range, Choices, Uniform, Beta, Normal, Enumeration, int, float, bool, str]:
```
rfatideh, line 351: We need to check if this conversion is needed. ref to check in ui-core: `modifierV2ToV3`

abhilash-kumar-nair:
Thanks @efredriksson-modelon, @rfatideh  This fix should be covered by this PR - https://github.com/modelon-community/impact-client-python/pull/343.

rfatideh:
https://github.com/modelon-community/impact-client-python/pull/343 will cover this

PR #345, opened 2026-06-15
feat: add get_extends method to Model entity

abhilash-kumar-nair:
**Description**
This PR adds an API to get all the models a given class extends.

**How to test**
import os
os.environ["IMPACT_PYTHON_CLIENT_EXPERIMENTAL"] ="true"
from modelon.impact.client import Client, ProjectType

client = Client(url="https://impact.modelon.cloud/", interactive=True)
model_path = "Unnamed.ex1_1_1"
ws = client.get_workspace("opo")
with ws.new_modeling_session() as session:
    model = session.get_model(model_path)
    print(model.get_extends())

### DIFF
--- a/modelon/impact/client/entities/model.py
+++ b/modelon/impact/client/entities/model.py
@@ -50,6 +50,11 @@ class ModelParameter:
 
 
 @dataclass
+class ExtendsClause:
+    class_name: str
+
+
+@dataclass
 class ExperimentDefinitionEntry:
     id: str
     name: str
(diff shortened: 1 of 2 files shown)


modelon/impact/client/entities/model.py
```diff
@@ -512,6 +512,20 @@ def get_source(self) -> str:
         modeling_sal = self._require_modeling_session("get_source")
         return modeling_sal().get_model_source(self._class_name)
 
+    @Experimental
+    def get_extends(self) -> List[str]:
```
efredriksson-modelon: Could use `get_extends_clauses`? It is slightly more verbose but I think it will read nicer. Just `get_extends` make you guess what is going on.

Also as you can have `extends QuiescentModelWithInheritance(gamma=0.3, delta=0.01);` in Modelica so I would suggest return `List[ExtendsClause]` so this introspection API can be easily be extended to in the future.

efredriksson-modelon (changes requested):

modelon/impact/client/entities/model.py
abhilash-kumar-nair: Good point. Yes that makes sense! :)

efredriksson-modelon (approved):

PR #346, opened 2026-06-16
refactor: extract FMU definition builder and simplify get_definition

abhilash-kumar-nair:
This MR adresses the comments in https://github.com/modelon-community/impact-client-python/pull/343

### DIFF
--- a/docs/source/modelon.impact.client.experiment_definition.rst
+++ b/docs/source/modelon.impact.client.experiment_definition.rst
@@ -44,6 +44,14 @@ modelon.impact.client.experiment\_definition.fmu\_based module
    :undoc-members:
    :show-inheritance:
 
+modelon.impact.client.experiment\_definition.from\_dict module
+--------------------------------------------------------------
+
+.. automodule:: modelon.impact.client.experiment_definition.from_dict
+   :members:
+   :undoc-members:
+   :show-inheritance:
+
 modelon.impact.client.experiment\_definition.model\_based module
 ----------------------------------------------------------------
 
--- a/docs/source/modelon.impact.client.experiment_definition.rst
+++ b/docs/source/modelon.impact.client.experiment_definition.rst
@@ -44,14 +44,6 @@ modelon.impact.client.experiment\_definition.fmu\_based module
    :undoc-members:
    :show-inheritance:
 
-modelon.impact.client.experiment\_definition.from\_dict module
---------------------------------------------------------------
-
-.. automodule:: modelon.impact.client.experiment_definition.from_dict
-   :members:
-   :undoc-members:
-   :show-inheritance:
-
 modelon.impact.client.experiment\_definition.model\_based module
 ----------------------------------------------------------------
 
--- a/modelon/impact/client/entities/experiment.py
+++ b/modelon/impact/client/entities/experiment.py
@@ -6,14 +6,16 @@ import warnings
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
 
-from modelon.impact.client.entities._initialize_from import _resolve_initialize_from
+from modelon.impact.client.entities._initialize_from import (
+    _resolve_extension_initialize_from,
+    _resolve_initialize_from,
+)
 from modelon.impact.client.entities.asserts import assert_variable_in_result
 from modelon.impact.client.entities.case import Case
 from modelon.impact.client.entities.custom_function import (
     CustomFunction,
     _build_custom_function,
 )
-from modelon.impact.client.entities.external_result import ExternalResult
 from modelon.impact.client.entities.interfaces.experiment import ExperimentReference
 from modelon.impact.client.entities.model import (
     Model,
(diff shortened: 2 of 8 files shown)


modelon/impact/client/entities/experiment.py
```diff
@@ -194,6 +196,62 @@ def label(self) -> Optional[str]:
...
+        for mod in modifiers.get("variables", [])
+    }
+    return SimpleFMUExperimentDefinition(
+        fmu=ModelExecutable(workspace_id, fmu_id, sal),
+        custom_function=custom_function,
+        solver_options=analysis.get("solverOptions", {}),
+        simulation_options=analysis.get("simulationOptions", {}),
+        simulation_log_level=analysis.get("simulationLogLevel", "WARNING"),
+        initialize_from=_resolve_initialize_from(workspace_id, sal, modifiers),
+    ).with_modifiers(modifiers=variable_modifiers)
```
efredriksson-modelon: This one does not support expansion? I think that it would make sense to have this one and `_build_simple_modelica_experiment_definition` be defined in the same file as they are so similar? The only difference between them is if you give 'fmu' or 'model' (+ compiler options if model). For example a local helper for
```
variable_modifiers = {
        mod["name"]: get_operator_from_dict(mod)
        for mod in base.get("modifiers", {}).get("variables", [])
}
```
would make sense.

modelon/impact/client/entities/experiment.py
```diff
@@ -669,52 +701,15 @@ def get_definition(self) -> ValidExperimentDefinitions:
...
-                    simulation_log_level=analysis.get("simulationLogLevel"),
-                    initialize_from=self._get_extension_initialize_from(
-                        extension["modifiers"]
-                    )
-                    if extension.get("modifiers")
-                    else None,
+            definition = _build_simple_fmu_experiment_definition(
+                base, custom_function, self._workspace_id, self._sal
+            )
+        if extensions_data:
```
efredriksson-modelon: I think it would make sense if the `_build` methods took `info` instead of `base` so they could add extensions themself.

Could even go so far of just having a single build function that does the `_get_workflow` check inline and return either definition type. Could have an assert for `get_experiment_definitions` that FMU is not returned as this should never happen given the API. So type matches `SimpleModelicaExperimentDefinition`.

efredriksson-modelon (changes requested):
I think this is an improvement 🧇 

What I would like to avoid is things like the FMU experiment missing expansion data. I think this becomes much simpler if the two builders live close to each other. Or, even is unified to a single function that returns `ValidExperimentDefinitions` and handle the entire dict -> entity part regardless of workflow.

modelon/impact/client/entities/experiment.py
abhilash-kumar-nair: Moved both builders into a new experiment_definition/from_dict.py so they sit next to each other, and pulled out the shared _get_variable_modifiers(modifiers) helper you suggested, both _build_modelica_definition and _build_fmu_definition use it now.

modelon/impact/client/entities/experiment.py
abhilash-kumar-nair: There's now a single _build_experiment_definition(info, custom_function, workspace_id, sal, model=None) that takes the full experiment dict, does the workflow check inline (base["model"].get("modelica")), attaches the extensions itself, and returns ValidExperimentDefinitions.

Experiment.get_definition is three lines now, and Model.get_experiment_definitions goes through the same function with assert isinstance(definition, SimpleModelicaExperimentDefinition), as you suggested. Side effect worth noting: it now passes item["experiment"] rather than just base, so saved definitions with extensions no longer lose them.

docs/source/modelon.impact.client.experiment_definition.rst
```diff
@@ -44,6 +44,14 @@ modelon.impact.client.experiment\_definition.fmu\_based module
...
    :show-inheritance:
 
+modelon.impact.client.experiment\_definition.from\_dict module
+--------------------------------------------------------------
+
+.. automodule:: modelon.impact.client.experiment_definition.from_dict
+   :members:
+   :undoc-members:
+   :show-inheritance:
+
```
efredriksson-modelon: Why are we documenting this module, we don't expect anyone to use it and is really only for internal dict/build helpers?

I would suggest:
 * removing this doc
 * rename _build_experiment_definition -> build_experiment_definition so we don't import private methods, but add to the doc-string of it that it is an 'internal' XXX.

Alternative we can rename:
 * from_dict -> _from_dict to show this is an internal module
 * make build_experiment_definition public. This might be more sane and probably how we though with the _initialize_from module, but we should change so methods we import from it are public.

efredriksson-modelon (approved):
Looks good 🍎 

Just one comment on the public/private naming that would be good to make a choice on for this PR.

docs/source/modelon.impact.client.experiment_definition.rst
abhilash-kumar-nair: Make sense! Thanks!

